PR #102, opened 2019-03-23
Rename Ethereum-related event services to `EthEventX`

### LINKED ISSUE #31: Rename events services related to ETH to `EthEvent` services
This is a pretty simple task! 

We have a few services that are used to watch Ethereum events. These are currently held in the [events](https://github.com/plasma-group/pigi/tree/master/packages/core/src/services/events) folder under `/src/services`. However, we *also* have a different [EventService](https://github.com/plasma-group/pigi/blob/master/packages/core/src/services/event.service.ts) that's used to shuttle events between services. That's a little confusing :-/.

We just need to rename `EventWatcherService` and `EventHandlerService` to `EthEventWatcherService` and `EthEventHandlerService`. Any imports of those services need to be updated too.

kfichter:

Fixes #31 

## Description

Renamed everything related to Ethereum events to `EthEventX` for clarity. Also went ahead and made those a subfolder of `services/eth` so it's clear where that all sits.


ben-chain:
All looks good/makes sense to me!

One question/idea: I noticed the pretty-print exports decimal representation.  One alternative might be to export as hex with `'0x' + value.toString(16)`, since values are often expressed as hex in ETH and otherwise.

Not sure if this has much impact or if we'd prefer decimal since we often give decimals in testing etc.  I was thinking of addresses in this case but perhaps those aren't `BigNum`s anyway--though maybe for debugging things like the tree where the address is concatenated with the `stateID` indexes it might be useful.

kfichter:
So actually the reasoning behind exporting as decimal was because pretty print is mainly used to debug range logic. That debugging was much harder when the numbers were in hex. Maybe we could do both? Print as hex and then decimal in parentheses.

packages/core/src/app.module.ts
```diff
@@ -10,13 +10,12 @@ import {
   OperatorService,
   ChainService,
   EthModule,
-  EventModule,
```
karlfloersch: Ah I see so you removed EventModule because it's now in the eth-event service. Seems reasonable!

karlfloersch:
> So actually the reasoning behind exporting as decimal was because pretty print is mainly used to debug range logic. That debugging was much harder when the numbers were in hex. Maybe we could do both? Print as hex and then decimal in parentheses.

Hmm yeah I kinda like hex representations personally. It feels nice to start thinking in hex more generally especially when dealing with a lot of hashes. That said I do see issues with hex not being user friendly so could very well be that we should just get into the habit of always printing our ranges in dec.... Idk I'm on the fence I guess with a slight leaning towards hex overall. That's an interesting idea to include both though don't know how messy/clean it looks in practice.

Idk that's my take but trust your judgement.

kfichter:
Made the above fixes. Merging!

PR #103, opened 2019-03-23
Added better exception handling

### LINKED ISSUE #79: Create more expressive error handling
Errors should be handled like they are in `@pigi/operator` using a base class that other exceptions extend. We should also make sure that error messages ask users to submit bug reports and that we log things!

kfichter:

Fixes #79 

## Description

Adds better exception handling. Primarily adds custom exceptions instead of `throw new Error('...')`. Also makes sure that errors extend a base class that asks users to report the bug if it's not expected.


kfichter:
Replaced by #174

Issue #72, opened 2019-03-21
Redesign how objects are inserted into the database


kfichter:
We're currently inserting objects into the database in a kinda silly way. This is especially true for arrays. We want to redesign the database insertions to really take advantage of the `key:value` store structure.

PR #105, opened 2019-03-23
Use better database prefixes

kfichter:

## Description

Starting to tackle #72... I'm trying to come up with a better way to handle inserting things into the database.

This PR makes the simple change of converting database prefixes into constants. For example, we previously used prefixes like so:

```
db.set(`prefixname:${key}`, value
```

Instead, we now have a `constants` file:

```
// constants.ts
export const DB_PREFIXES = {
  PREFIX_NAME: '00',
}
```

so the above example becomes:

```
db.set(`${PREFIX_NAME}:${key}`, value)
```

It's certainly cleaner and ensures that we don't accidentally typo up a prefix.


packages/core/src/constants.ts
```diff
@@ -6,3 +6,22 @@ export const CONFIG = {
...
+  EXITABLE_ENDS: '07',
+  FINALIZED_EXITS: '08',
+  EXITED_RANGES: '09',
+  FAILED_TRANSACTION_IMPORTS: '10',
+  SEEN_EVENTS: '11',
+  LATEST_BLOCK: '12',
+  LATEST_STATE: '13',
+  LAST_LOGGED_BLOCK: '14',
+  LAST_SYNCED_BLOCK: '15',
+}
```
karlfloersch: These should probably be a single byte value to avoid weird encoding / decoding unknown behavior. So these DBs store things as bytes so if we use strings to `put` then we may have some weird thing where one key is encoded `12ab` and the other key `12ab1b`, and that would be bad!

So probably best to use something like:
```
SEEN_EVENTS: Buffer.from([255]),
```
That way we know it's a single byte and we know there won't be some unintended overlap.

karlfloersch: Oh another thing you can do is you can create an object which allows you to "register" prefixes, and it does the Buffer management on the backend. I'd say that looking at this will be helpful: 

https://github.com/bcoin-org/bdb

kfichter:
Closing as this PR in no longer relevant!

PR #108, opened 2019-03-25
Update how we represent state to match spec

kfichter:

## Description
This PR updates `@pigi/utils` with three new elements, `Transaction`, `StateUpdate`, and `StateObject`. The basic structure of these elements is as follows:

```
StateObject {
  parameters: bytes
  predicate: address
}

StateUpdate {
  start: uint
  end: uint
  block: uint
  chain: address
  newState: StateObject
}

Transaction {
  stateUpdate: StateUpdate
  witness: bytes
}
```


ben-chain:
Hmmmm so this is a little tricky--what do you consider the witness here?  Presumably it's the signature of the owner?  If this is the thing you send to the operator it's notably not the deprecation witness, because that will also have the inclusion proof in it which doesn't exist.  It's almost like an updateability witness, distinct from the deprecation witness.

kfichter:
> Hmmmm so this is a little tricky--what do you consider the witness here? Presumably it's the signature of the owner? If this is the thing you send to the operator it's notably not the deprecation witness, because that will also have the inclusion proof in it which doesn't exist. It's almost like an updateability witness, distinct from the deprecation witness.

Yes it's a transaction witness, not the deprecation witness.

ben-chain:
Great, makes sense--I think picking a name to distinguish it as such might be a good call.

Other thing I noticed--`chain` here is the plasma contract right?  Might be worth calling it `plasmaContract` so that it's not confused with the address of the commitment contract.

ben-chain:
Otherwise, looks good to me :)

packages/utils/src/state/state-object.ts
```diff
@@ -0,0 +1,31 @@
...
+ * blocks of our plasma chain design.
+ */
+export class StateObject {
+  public predicate: string
+  public parameters: string
+
+  constructor(args: StateObjectData) {
+    this.predicate = args.predicate
+    this.parameters = args.parameters
+  }
```
karlfloersch: Curious why break it into StateObjectData & StateObject instead of just passing the args directly into the constructor. This is clearly very intentional so I'm just curious what I'm missing

kfichter: Personal preference. I like the following construction:

```
const StateObject = new StateObject({
  predicate: ‘0x...’,
  parameters: ...
})
```

Main advantage is that it’s clear what you’re passing into the constructor w/o having to look at the constructor definition.

kfichter:
> Great, makes sense--I think picking a name to distinguish it as such might be a good call.
> 
> Other thing I noticed--`chain` here is the plasma contract right? Might be worth calling it `plasmaContract` so that it's not confused with the address of the commitment contract.

Made these renaming changes :+1:

Issue #99, opened 2019-03-22
Add `iterator` to DB service


karlfloersch:
**Is your feature request related to a problem? Please describe.**
Right now we only have a function called `findNextKey` which simply retrieves the next key based on some prefix. However, one should be able to first find a key based on a prefix, and then iterate over the DB starting from that point efficiently. Repeatedly calling `findNextKey` would not be efficient. We should also be able to do this in reverse.

**Describe the solution you'd like**
Create a function in `base-db.provider.ts` called `seek()` which has similar functionality to [seek() in LevelDown](https://github.com/Level/leveldown#iteratorseekkey). There should also be a way to run seek in reverse.

**Describe alternatives you've considered**
None I guess. Would be happy to hear more ideas.

PR #114, opened 2019-03-26
Add LevelDB support in DB service

### LINKED ISSUE #113: Add LevelDB support to DB service
Currently the DB service does not support leveldb... that's no good! We need leveldb for the operator.

Add a class which is similar to EphemDB but instead uses a LevelDB backend.

karlfloersch:

Fixes #113 by adding a LevelDB class which creates a LevelDB instance & then implements `get`, `put`, `del`, `iterator`, etc. It does not yet implement `seek`.

## Description
This PR begins the process of adding LevelDB support and moving our DB abstraction layer to something closer to what we want in the long term. There are some shortcomings though of this current PR. These are:

1. There was no reimplementation of `push()` inside of core, and instead I just changed it to `put()`. I didn't feel so bad about doing this considering a lot of other tests are failing anyway, so even if that broke something it seems this whole codebase is very much WIP.
2. `batch()` and `seek()` are not actually implemented properly in `EphemDB`. Instead they are just clones of `batch()` -> `bulkPut()` and `seek()` -> `findNextKey()`
3. `batch()` was not tested for LevelDB.
4. `seek()` was not implemented for LevelDB

### Small Misc Change
I removed the prefix from `wsrun` because I had some trouble getting colored text to show up when the prefix was active. I found this solution a day or two ago suggested in a GH issue, but now I can't find where I found this solution. I only added it to `test` because that was the one that really needed it, but deff no problems with removing the prefix all together.


packages/core/src/services/db/backends/ephem-db.provider.ts
```diff
@@ -88,9 +93,9 @@ export class EphemDBProvider implements BaseDBProvider {
    * Should be more efficient than simply calling `set` repeatedly.
    * @param objects A series of objects to put into the database.
    */
-  public async bulkPut(objects: DBObject[]): Promise<void> {
+  public async batch(objects: DBObject[]): Promise<void> {
```
kfichter: I believe this should be `public async batch(ops: DBOperation[]): Promise<void> {`

packages/core/src/services/db/backends/ephem-db.provider.ts
```diff
@@ -88,9 +93,9 @@ export class EphemDBProvider implements BaseDBProvider {
...
     for (const object of objects) {
-      await this.set(object.key, object.value)
+      await this.put(object.key, object.value)
```
kfichter: Should we check the operation here (`put` or `del`)?

packages/core/src/services/db/backends/base-db.provider.ts
```diff
@@ -3,6 +3,12 @@ export type DBValue = string | object | number | boolean
 export type DBResult = DBValue | DBValue[]
 
 export interface DBObject {
+  key: Buffer | string
+  value: DBValue
+}
+
+export interface DBOperation {
+  type: string
```
kfichter: Probably want to use `type: 'put' | 'del'`

packages/core/src/services/db/backends/ephem-db.provider.ts
```diff
@@ -103,6 +108,19 @@ export class EphemDBProvider implements BaseDBProvider {
...
+  }
+
+  /**
+   * Accepts a string or Buffer and returns a string representation of
+   * the input.
+   * @param value
+   * @returns the value as a string
+   */
+  private convertToString(value: Buffer | string) {
+    if (typeof value === 'string')
```
kfichter: `prettier` will probably complain about missing brackets here.

packages/core/src/services/db/backends/ephem-db.provider.ts
```diff
@@ -18,7 +18,8 @@ export class EphemDBProvider implements BaseDBProvider {
    * @param fallback A fallback value if the key doesn't exist.
    * @returns the stored value or the fallback.
    */
-  public async get<T>(key: string, fallback?: T): Promise<T | DBResult> {
+  public async get<T>(_key: Buffer | string, fallback?: T): Promise<T | DBResult> {
```
kfichter: Hmm I'd vote to keep the parameter as `key` instead of `_key`.

packages/core/src/services/db/backends/level-db.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  /**
+   * Finds the next key after a given key.
+   * @param key The key to start searching from.
+   * @returns the next key with the same prefix.
+   */
+  public async seek(key: string): Promise<string> {
+    return 'NOT IMPLEMENTED'
```
kfichter: Probably better to throw an error instead of returning. Maybe `throw new Error('Method not implemented')`.

packages/core/src/services/db/backends/ephem-db.provider.ts
karlfloersch: Whoops totally right -- updating

packages/core/src/services/db/backends/base-db.provider.ts
karlfloersch: Oooo that's very cool

packages/core/src/services/db/backends/ephem-db.provider.ts
karlfloersch: For sure we should. However, I just didn't want to break the dependencies for ephemDB & so I kept the implementation of batch as being what `bulkPut` used to be. Basically I'm trying to make minimal changes which would allow me to integrate the operator using the DB service

packages/core/src/services/db/backends/ephem-db.provider.ts
karlfloersch: You're right! There's a bunch of linting errors -- need to update my editor's linter to follow prettier instead of whatever it's using now

kfichter: You can also run `yarn fix` in the root to fix issues automatically :+1:

packages/core/src/services/db/backends/base-db.provider.ts
karlfloersch: Hmm so does this mean the type is `a string which equals 'put' or 'del'`? Or does this mean the put or del functions? I was having trouble getting this to work except by using either `string` or `Function`

packages/core/src/services/db/backends/ephem-db.provider.ts
karlfloersch: Ah so the reason I was doing this is I can't reassign the parameter & didn't feel like using a different variable name for the key that's used in the function. For instance I could use `const stringifiedKey = this.convertToString(key)` and then use the stringifiedKey.

karlfloersch:
Addresses #99

kfichter:
@karlfloersch this may be outdated now that we have the new [standard database implementation](https://github.com/plasma-group/pigi/tree/3fbe2e9a96e21a51d47744aa220287ce9c9a8051/packages/core/src/app/common/db) that supports any [abstract-leveldown](https://github.com/Level/abstract-leveldown/#abstractleveldown_iteratoroptions) compliant back-end.

PR #170, opened 2019-04-08
Cleaned up PlasmaRegistry.vy

kfichter:

## Description
Made some minor cleanup changes to `PlasmaRegistry`. Mainly added some comments, docstrings, and used `__init__` instead of a separate initialization function. Also added a mapping of `name => address` so we can easily query the registry instead of looking at events!


ben-chain:
Looks good to me!  Note that I have a contract deployment function built out for `ethereum.ts` so will add an issue to test this.

PR #182, opened 2019-04-14
Added basic proof generation and supporting methods

### LINKED ISSUE #28: Create a new `ProofGenerationService`
We want all proof-related functionality to be placed into a single module, "ProofModule". We have two distinct main functionalities, proof verification and proof generation. These two functionalities should each be placed into their own services. 

Luckily, the current [ProofService](https://github.com/plasma-group/pigi/blob/master/packages/core/src/services/chain/proof-service.ts) is the equivalent of `ProofVerificationService`. We just need to implement a new service, `ProofGenerationService`, that, given a transaction, will generate the proof necessary to validate the transaction.

kfichter:

Fixes #28, 60

## Description
Adds proof generation logic to `HistoryManager`. Adds lots of methods to `ChainDB` to support proof generation.

This is a WIP and needs to be heavily tested!

### DIFF
--- a/packages/core/src/app/pg-core/db/chain-db.ts
+++ b/packages/core/src/app/pg-core/db/chain-db.ts
@@ -21,8 +21,6 @@ const computeParent = (leftChild: string, rightChild: string): string => {
   return
 }
 
-// TODO: Maybe this should go in a utils file?
-
 /**
  * Checks whether a node is a left child.
  * @param nodeIndex Node index to check
@@ -106,18 +104,18 @@ const getMerkleSiblingIndices = (
   return siblingIndices
 }
 
-// TODO: Where should this sit? Probably OK to be here.
 const KEYS = {
   BLOCKS: new BaseKey('b', ['uint32']),
   DEPOSITS: new BaseKey('d', ['uint256']),
   TRANSACTIONS: new BaseKey('t', ['uint32', 'uint256']),
   TREE_HEIGHTS: new BaseKey('h', ['uint32']),
   TREE_NODES: new BaseKey('n', ['uint32', 'uint32']),
+  LEAF_INDICES: new BaseKey('l', ['uint32', 'uint256']),
 }
 
 /**
- * Basic ChainDB implementation that provides a
- * nice interface to the chain database.
+ * Basic ChainDB implementation that provides a nice interface to the chain
+ * database.
  */
 export class PGChainDB implements ChainDB {
   /**
@@ -346,8 +344,14 @@ export class PGChainDB implements ChainDB {
     blockNumber: number,
     leafIndex: number
   ): Promise<void> {
-    // TODO: Figure out how to compute the leaf indices of other transactions.
-    const otherLeafIndices: number[] = []
+    // Get the leaf indices of all other transactions in the same block.
+    const otherLeafIndices: number[] = (await this.getAllLeafIndices(
+      blockNumber
+    )).filter((otherLeafIndex) => {
+      return otherLeafIndex !== leafIndex
+    })
+
+    // Need the tree height to compute sibling/parent nodes.
     const treeHeight = await this.getBlockTreeHeight(blockNumber)
 
     // We don't want to delete any nodes that other transactions still need.
@@ -384,9 +388,8 @@ export class PGChainDB implements ChainDB {
   public async getInclusionProof(
     transaction: Transaction
   ): Promise<InclusionProof> {
-    // TODO: Figure out how to compute the leaf index?
     const blockNumber = transaction.block
-    const leafIndex = 0
+    const leafIndex = await this.getLeafIndex(transaction)
     const treeHeight = await this.getBlockTreeHeight(blockNumber)
 
     // Get the list of siblings up the tree.
@@ -460,6 +463,51 @@ export class PGChainDB implements ChainDB {
     const key = KEYS.TRANSACTIONS.encode([transaction.block, end])
     const value = encode(transaction)
     await this.db.put(key, value)
+  }
+
+  /**
+   * Sets the leaf index for a given transaction.
+   * @param transaction Transaction to set.
+   * @param leafIndex Leaf index for the transaction.
+   */
+  public async addLeafIndex(
+    transaction: Transaction,
+    leafIndex: number
+  ): Promise<void> {
+    const end = getTransactionRangeEnd(transaction)
+    const key = KEYS.LEAF_INDICES.encode([transaction.block, end])
+    const value = Buffer.allocUnsafe(4)
+    value.writeUInt32BE(leafIndex, 0)
+    await this.db.put(key, value)
+  }
+
+  /**
+   * Gets the leaf index for a transaction.
+   * @param transaction Transaction to query.
+   * @returns the leaf index for that transaction.
+   */
+  public async getLeafIndex(transaction: Transaction): Promise<number> {
+    const end = getTransactionRangeEnd(transaction)
+    const key = KEYS.LEAF_INDICES.encode([transaction.block, end])
+    const value = await this.db.get(key)
+    return value.readUInt32BE(0)
+  }
+
+  /**
+   * Gets all leaf indices for a given block.
+   * @param blockNumber Block to get leaf indices for.
+   * @returns all leaf indices for that block.
+   */
+  public async getAllLeafIndices(blockNumber: number): Promise<number[]> {
+    const iterator = this.db.iterator({
+      gte: KEYS.LEAF_INDICES.min([blockNumber]),
+      lte: KEYS.LEAF_INDICES.max([blockNumber]),
+    })
+
+    const values = await iterator.values()
+    return values.map((value) => {
+      return value.readUInt32BE(0)
+    })
   }
 
   /**
--- a/packages/core/src/app/pg-core/db/chain-db.ts
+++ b/packages/core/src/app/pg-core/db/chain-db.ts
@@ -22,6 +22,55 @@ const computeParent = (leftChild: string, rightChild: string): string => {
   return
 }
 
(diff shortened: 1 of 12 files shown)


packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
- * Basic ChainDB implementation that provides a
- * nice interface to the chain database.
+ * Checks whether a node is a left child.
+ * @param nodeIndex Node index to check
+ * @returns `true` if the node is a left child, `false` otherwise.
  */
-export class PGChainDB implements ChainDB {
-  private statedb: KeyValueStore
-  private historydb: KeyValueStore
+const isLeftChild = (nodeIndex: number): boolean => {
```
kfichter: @karlfloersch Do you think it makes sense for these binary tree utilities to live in a separate utils file? Might make them easier to test.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+import {
+  BaseKey,
+  encode,
+  decode,
+  bnToUint256,
+  getTransactionRangeEnd,
+} from '../../common'
+
+// TODO: Where should this stuff sit? Should it be customizeable?
+const computeParent = (leftChild: string, rightChild: string): string => {
```
kfichter: @karlfloersch wasn't exactly sure how to handle this. We somehow need to provide the parent hashing function to this class.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+    siblingIndices.push(siblingIndex)
+
+    // Go on to the parent.
+    nodeIndex = getParentIndex(nodeIndex)
+  }
+
+  return siblingIndices
+}
+
+const KEYS = {
```
kfichter: Do these keys make sense to people?

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+    return transactions
+  }
+
+  /**
+   * Queries deposits in a given range.
+   * @param start Start of the range to query.
+   * @param end End of the range to query.
+   * @returns a list of deposits for the range.
+   */
+  public async getDeposits(start: BigNum, end: BigNum): Promise<Transaction[]> {
```
kfichter: Deposits are separated from Transactions because you can't have multiple deposits on the same block. No need to store deposits on a per-block basis. Can just store them all under the same key and identified by `range.end`.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+  }
+
+  /**
+   * Queries a specific stored Merkle tree node. Attempts to compute the node
+   * from its children if the node doesn't exist.
+   * @param blockNumber Block number to pull the node from.
+   * @param nodeIndex Index of the specific node to query.
+   * @returns the node or `null` if the node cannot be found.
+   */
+  public async getMerkleTreeNode(
```
kfichter: Here's where things get a little complex - this is a recursive method for seeing if we have a given Merkle tree node. If we don't have the node itself, this method checks if we can generate it from some set of children.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+  ): Promise<void> {
+    // Don't do anything if we already have the node.
+    if ((await this.getMerkleTreeNode(blockNumber, nodeIndex)) !== null) {
+      return
+    }
+
+    // Smart parent checks. We don't need to store the parent as long as we
+    // have this node's sibling. Parent can simply be re-generated later on.
+    // Skip if this is the root node because it doesn't have a sibling.
+    if (nodeIndex !== 0) {
```
kfichter: If we're adding a node that has a sibling, then we no longer need to keep the node's parent since we can generate it from the two children.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+    if (value !== null) {
+      // Delete the node.
+      await this.db.del(key)
+
+      // Figure out if this node had a sibling. If so, we're going to want to
+      // compute the parent and insert it since we'd be losing that info.
+      const siblingIndex = getSiblingIndex(nodeIndex)
+      const sibling = await this.getMerkleTreeNode(blockNumber, siblingIndex)
+
+      // We have a sibling, so we're going to compute the parent and insert it.
```
kfichter: We had a sibling, meaning we could've generated the parent. If we simply deleted this node, then we'd lose that parent information. We therefore want to insert the parent again so we don't lose any information other than what we specifically wanted to remove.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+      return otherLeafIndex !== leafIndex
+    })
+
+    // Need the tree height to compute sibling/parent nodes.
+    const treeHeight = await this.getBlockTreeHeight(blockNumber)
+
+    // We don't want to delete any nodes that other transactions still need.
+    // Compute the siblings for *all* other stored transactions so we're not
+    // deleting anything critical.
+    let allOtherSiblingIndices: number[] = []
```
kfichter: Makes sure we don't delete any nodes that *other* transactions in the same block still need.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+      }
+    }
+  }
+
+  /**
+   * Creates an inclusion proof for a given transaction.
+   * @param transaction Transaction to create a proof for.
+   * @returns the inclusion proof for that transaction.
+   */
+  public async getInclusionProof(
```
kfichter: Not sure how this actually plays with the Merkle sum tree. Is this method independent of the underlying Merkle tree?

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+    const value = Buffer.from(blockHash)
+    await this.db.put(key, value)
+  }
+
+  /**
+   * Inserts the tree height for a given block.
+   * @param blockNumber Block number to add tree height for.
+   * @param treeHeight Height of the tree.
+   */
+  public async addBlockTreeHeight(
```
kfichter: We need to store the height of the tree for each block. Otherwise we wouldn't be able to determine if a given index is valid.

packages/core/src/app/pg-core/db/chain-db.ts
```diff
@@ -1,24 +1,523 @@
...
+    const value = encode(transaction)
+    await this.db.put(key, value)
+  }
+
+  /**
+   * Sets the leaf index for a given transaction.
+   * @param transaction Transaction to set.
+   * @param leafIndex Leaf index for the transaction.
+   */
+  public async addLeafIndex(
```
kfichter: Need to store the leaf indices of each transaction. @karlfloersch is this OK or do you think there's a cleaner way to store this?

packages/core/src/app/pg-core/state/history-manager.ts
```diff
@@ -1,48 +1,112 @@
...
+   * Assumes that the recipient has access to public information (e.g. plasma
+   * blocks).
    * @param transaction Transaction to generate a proof for.
    * @returns the transaction proof.
    */
   public async getTransactionProof(
     transaction: Transaction
   ): Promise<TransactionProof> {
-    return
+    /**
```
kfichter: @karlfloersch would love your feedback on this. Does the process look right?

packages/core/src/app/pg-core/db/chain-db.ts
karlfloersch: Ah yes I believe you have to store the index of the transaction. There might be a cleaner way but I can't think of it off the top of my head.

One weird note is that I actually was using the start position as the key, seeking to the end of the range I care about, and then doing a reverse iteration. Tbh I'm not sure why I did that instead of storing it by the end, seeking with the start, and using a forward iteration. However, just note that's what I did and there *may* be a reason but I'm not sure.

packages/core/src/app/pg-core/state/history-manager.ts
```diff
@@ -1,48 +1,112 @@
...
+     * part of the block it's supposed to be in.
+     */
+    for (const tx of transactions) {
+      const inclusionProof = await this.db.getInclusionProof(tx)
+
+      proof.push({
+        transaction: tx,
+        inclusionProof,
+      })
+    }
```
karlfloersch: I like how you separated getting the transactions from getting the inclusion proofs. Feels pretty clean.

packages/core/src/app/pg-core/state/history-manager.ts
karlfloersch: Everything looks good! My only concern is whether `getTransactions()` handles empty blocks properly

packages/core/src/app/pg-core/db/chain-db.ts
karlfloersch: Hmm I've gone back and forth about this. The ChainDB is quite large and so it would be nice to break these out into another file. Plus they could be useful for other implementations of ChainDB that use merkle trees. That said, if it's in another file I do feel that file should live very close to this ChainDB file so it doesn't just get added to a massive utils file with random stuff in it.

karlfloersch:
This looks pretty great and I am excited to see tests!

PR #197, opened 2019-05-18
Fix ownership tx & add tests

karlfloersch:
Fix encoding & add a few tests


karlfloersch:
Once again wrong repository...

PR #199, opened 2019-05-25
Add new contracts from EthNY

### LINKED ISSUE #198: Use Waffle to test contracts
Our smart contracts use a custom testing framework & vyper. This is not great considering we want to use solidity & reduce the maintenance requirements of our project. 

Move contracts to Solidity and use waffle framework for testing.

karlfloersch:

Fixes #198

## Description
This pull request revamps the `contracts` package to use Waffle & also integrates the contract @alexanderattar wrote during the EthNY hackathon. 

The only regret is this PR also squashes some bugs in `core`. However, arguably this is not a problem considering core's functionality is already quite out of date & needs to be updated.


karlfloersch:
@ben-chain I just edited the PR so that instead of squashing the bugs in core, I just disabled the build & test script as I did for PR https://github.com/plasma-group/pigi/pull/201

packages/contracts/contracts/BasicTokenMock.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.5.1;
```
kfichter: What's this contract for?

packages/contracts/contracts/Commitment.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "openzeppelin-solidity/contracts/math/Math.sol";
```
kfichter: Is this import used?

packages/contracts/contracts/BasicTokenMock.sol
karlfloersch: Ah am using this contract as the ERC20 the deposit contract.

packages/contracts/contracts/Commitment.sol
karlfloersch: Nope! Got rid of it because can't think of when it would be.

alexanderattar: I actually don't think the Commitment.sol contract is necessary anymore. Initially we were adding the verifyInclusion method on this and use it within Deposit.sol, but ended up just putting the method directly on the Deposit contract.

ben-chain: Yep--but that was more for hackathon purposes, we would have needed to pull into an external contract eventually :)

alexanderattar: Yeah that's a good point. Probably just better to keep and iterate on it then.

PR #243, opened 2019-06-10
Add Basic Deposit Contract

### LINKED ISSUE #214: Implement basic deposit contract
We need to implement a basic [deposit contract](http://spec.plasma.group/en/latest/src/02-contracts/deposit-contract.html) that allows users to deposit ETH or ERC-20 tokens. Requires that we first implement a [commitment contract](http://spec.plasma.group/en/latest/src/02-contracts/commitment-contract.html).

karlfloersch:

Fixes #214

## Description
This pull request adds a basic `Deposit.sol` contract with testing for:

- depositing
- starting checkpoints
- starting exits
- deprecating exits
- deleting outdated exits with a checkpoint
- challenging checkpoints
- removing challenges
- deposited ranges logic


packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+        bytes32 checkpointId = getCheckpointId(checkpoint);
+        CheckpointStatus memory status = CheckpointStatus(
+            {challengeableUntil: block.number + CHALLENGE_PERIOD, outstandingChallenges: 0});
+        checkpoints[checkpointId] = status;
+        // Emit an event which informs us that the checkpoint was finalized
+        emit CheckpointFinalized(checkpointId);
+        emit LogCheckpoint(checkpoint);
+    }
+
+    function extendDepositedRanges(uint256 _amount) public {
```
kfichter: Assuming this shouldn't be public.

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+            // Case 2: We are extending the old range (deleting the old range and making a new one with the total length)
+            delete depositedRanges[oldEnd];
+            newStart = oldStart;
+        }
+        // Set the newEnd to the totalDeposited plus how much was deposited
+        uint256 newEnd = totalDeposited + _amount;
+        // Finally create the range!
+        newRange = dt.Range(newStart, newEnd);
+        // Store the result
+        depositedRanges[newRange.end] = newRange;
```
kfichter: Could probably combine these two lines.

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+        // Increment total deposited now that we've extended our depositedRanges
+        totalDeposited += _amount;
+    }
+
+    function removeDepositedRange(dt.Range memory range, uint256 depositedRangeId) public {
```
kfichter: Assume this also shouldn't be public.

kfichter: Probably worth adding more comments, method is a little confusing.

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+    function challengeCheckpoint(Challenge memory _challenge) public {
+        bytes32 challengedCheckpointId = getCheckpointId(_challenge.challengedCheckpoint);
+        bytes32 challengingCheckpointId = getCheckpointId(_challenge.challengingCheckpoint);
+        bytes32 challengeId = getChallengeId(_challenge);
+        // Verify that the challenge may be added
+        require(exitExists(challengingCheckpointId), "Challenging exit must exist");
+        require(checkpointExists(challengedCheckpointId), "Challenged checkpoint must exist");
+        require(intersects(_challenge.challengedCheckpoint.subrange, _challenge.challengingCheckpoint.subrange), "Challenge ranges must intersect");
+        require(_challenge.challengingCheckpoint.stateUpdate.plasmaBlockNumber < _challenge.challengedCheckpoint.stateUpdate.plasmaBlockNumber, "Challenging cp after challenged cp");
+        require(challenges[challengeId] != true, "Challenge must not already exist");
```
kfichter: Can just `require(!challenges[challengeId], ...)`

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+        checkpoints[challengedCheckpointId].outstandingChallenges += 1;
+        challenges[challengeId] = true;
+    }
+
+    function removeChallenge(Challenge memory _challenge) public {
+        bytes32 challengedCheckpointId = getCheckpointId(_challenge.challengedCheckpoint);
+        bytes32 challengingCheckpointId = getCheckpointId(_challenge.challengingCheckpoint);
+        bytes32 challengeId = getChallengeId(_challenge);
+        // Verify that the challenge may be added
+        require(challenges[challengeId] == true, "Challenge must exist");
```
kfichter: Can just `require(challenges[challengeId], ...)`

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+        delete exitRedeemableAfter[checkpointId];
+    }
+
+    function deleteOutdatedExit(dt.Checkpoint memory _exit, dt.Checkpoint memory _newerCheckpoint) public {
+        bytes32 outdatedExitId = getCheckpointId(_exit);
+        bytes32 newerCheckpointId = getCheckpointId(_newerCheckpoint);
+        require(intersects(_exit.subrange, _newerCheckpoint.subrange), "Exit and newer checkpoint must overlap");
+        require(_exit.stateUpdate.plasmaBlockNumber < _newerCheckpoint.stateUpdate.plasmaBlockNumber, "Exit must be before a checkpoint");
+        require(checkpoints[newerCheckpointId].outstandingChallenges == 0, "Newer checkpoint must have no challenges");
+        require(checkpoints[newerCheckpointId].challengeableUntil < block.number, "Newer checkpoint must no longer be challengeable");
```
kfichter: Maybe worth turning these two calls into helper functions since they're used repeatedly.

packages/contracts/contracts/Commitment.sol
```diff
@@ -1,22 +1,12 @@
...
+import {DataTypes as dt} from "./DataTypes.sol";
 
-  struct StateUpdate {
-    Range range;
-    StateObject stateObject;
-    address plasmaContract;
-    uint256 plasmaBlockNumber;
-  }
-    function verifyInclusion(StateUpdate memory _stateUpdate, bytes memory _inclusionProof) public returns (bool) {}
+contract Commitment {
```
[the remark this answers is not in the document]
ben-chain: CommitmentChain?

packages/contracts/contracts/DataTypes.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
```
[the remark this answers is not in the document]
ben-chain: Believe so--it's doing nested encoding currently, I think the idea was to flatten down the line once everything is finalized.

packages/contracts/contracts/Deposit.sol
```diff
@@ -1,234 +1,281 @@
...
+        dt.Checkpoint memory checkpoint = dt.Checkpoint({
+            stateUpdate: stateUpdate,
+            subrange: depositRange
+        });
+        // Extend depositedRanges & increment totalDeposits
+        extendDepositedRanges(_amount);
+        // Calculate the checkpointId and add it to our finalized checkpoints
+        bytes32 checkpointId = getCheckpointId(checkpoint);
+        CheckpointStatus memory status = CheckpointStatus(
+            {challengeableUntil: block.number + CHALLENGE_PERIOD, outstandingChallenges: 0});
```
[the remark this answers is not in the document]
ben-chain: No, you're correct, otherwise you'd have to download exclusion proofs--my suggestion in the spec was to do ``block.number - 1`` since we check for deletion with ``==0``.  Will change if that makes sense.  Great catch!

packages/contracts/contracts/Deposit.sol
ben-chain: Yeah this was left public for testing, what do you think the best way to deal with testing is?

kfichter: Ideally you shouldn't need to test private methods (tests should cover things a user is allowed to do). You'd instead want to add extra tests to whatever method makes use of this method.

packages/contracts/contracts/Commitment.sol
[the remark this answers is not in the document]
kfichter: That works.

packages/contracts/contracts/Deposit.sol
karlfloersch: Maybe another option would be pulling out this functionality into another contract with public variables & testing that. Then we can import that contract & store a reference to it and keep that reference private

ben-chain:
Added Kelvin's changes minus the pure function business, we can pull out as a library later. ([seems like standard practice](https://ethereum.stackexchange.com/questions/38158/how-to-unit-test-contract-private-and-internal-functions))

PR #275, opened 2019-06-12
Added keystore interface

### LINKED ISSUE #255: Determine correct format for keystore objects
[Our format for keystore objects](http://spec.plasma.group/en/latest/src/05-client-architecture/wallet-db.html#keystore) was taken from documentation for [go-ethereum](https://github.com/ethereum/go-ethereum). We need to figure out if this keystore format is correct and/or if there's a specification for the keystore format somewhere online. Once we know, we should update the specification accordingly.

kfichter:

Fixes #255 

## Description
Found the correct format for keystore files!


karlfloersch:
@kfichter  looks like this was included inside of https://github.com/plasma-group/pigi/pull/277 -- is that right ?

PR #276, opened 2019-06-12
235 - Implementing StateManager.executeTransaction(...)

### LINKED ISSUE #235: Implement executeTransaction
```
async function executeTransaction(
  transaction: Transaction
): Promise<{ stateUpdate: StateUpdate, validRanges: Range[] }>
```

Description:
Executes a `transaction`_ against the current verified state and returns the resulting state update.

Transactions reference a range of state objects that they’re attempting to modify. It’s possible that we only have the full valid history for some of the referenced state objects but not others. This behavior is allowed by construction. As a result, this method also returns the list of ranges over which the transaction can be considered “valid”.

For example, we may have a valid history for the ranges (0, 50) and a transaction that sends (0, 100). We can assert that the transaction is valid for (0, 50) but cannot make the same assertion for (50, 100).

Parameters:
transaction - Transaction: `Transaction`_ to execute against the verified state.
Returns

`Promise<{ stateUpdate: StateUpdate, validRanges: Range[] }>`: The `StateUpdate` created as a result of the transaction and the list of ranges over which the state update has been validated.

willmeister:

Fixes #235 

## Description

* Adds StateManager and all of the data types it depends on
* Adds various dependencies (PluginManager, PredicatePlugin, StateDB, etc.)
* Implements the `StateManager.executeTansaction(...)` method, leaning on mocked dependency interfaces for now


packages/core/src/interfaces/common/utils/range-store.interface.ts
```diff
@@ -0,0 +1,11 @@
+
+import BigNum = require('bn.js')
+
+export interface Range {
+  start: BigNum
+  end: BigNum
+}
```
karlfloersch: You can import a range interface here -- https://github.com/plasma-group/pigi/blob/a1611bd23b13f9d70d246774a4900b36987ecfd6/packages/utils/src/interfaces/data-types.ts

Also we need to reintroduce these data-types which got deleted & were reintroduced here: https://github.com/karlfloersch/pigi/tree/master/packages/utils/src/data-types

karlfloersch: We could also put this stuff in core potentially just to keep the code together

willmeister: Will do. Which `Transaction` should I run with, unless they're supposed to be distinct?
* One: https://github.com/plasma-group/pigi/blob/master/packages/utils/src/interfaces/data-types.ts#L18
* Two: https://github.com/plasma-group/pigi/blob/master/packages/core/src/interfaces/common/utils/state.interface.ts#L17

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,119 @@
...
+      } else if (result.stateUpdate !== computedState) {
+        throw new Error(`State transition resulted in two different states: ${result.stateUpdate} and 
+          ${computedState}. Latter differed from former at range ${result.validRanges.pop()}.`)
+      }
+    })
+
+    return result;
+  }
+
+  public ingestHistoryProof(historyProof: HistoryProof): Promise<void> {
```
kfichter: So generally (... and @karlfloersch can comment here too since we've discussed before ...) I'm a fan of taking a snapshot of the local state when verifying history proofs instead of directly applying the proof elements to the database.

My reasoning:
1. It's much faster - you're not making repeated reads/writes.
2. It's easier to revert - you don't have to go back and erase data if the proof turns out to be invalid.
3. It's possible to check multiple proofs in parallel - you only need to "lock" a specific range when you're writing to the database after checking the proof.

Let me know what you think!

packages/core/src/interfaces/common/utils/range-store.interface.ts
karlfloersch: This is a great question. [The first one you listed](https://github.com/plasma-group/pigi/blob/master/packages/utils/src/interfaces/data-types.ts#L18) is the most up-to-date; however, the `method_id` should be optional. As you're starting out with the ownership predicate, just skip the `method_id` because in this case there's only one method, `send()`.

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,119 @@
...
+    const result = {
+      stateUpdate: undefined,
+      validRanges: []
+    }
+
+    if ( ! this.validateTransaction(transaction)) {
+      return result
+    }
+
+    // TODO: Check that transaction.block > SyncManager.getLastSyncedBlock()
```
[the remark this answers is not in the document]
willmeister: Good call. There's already a check in the for loop under this, so I just removed the TODO.

packages/core/src/interfaces/client/plugin-manager.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+   */
+  loadPlugin(address: string, path: string): Promise<PredicatePlugin>
+
+  /**
+   * Gets the PredicatePlugin associated with the provided address, if one exists.
+   *
+   * @param address the address of the PredicatePlugin
+   * @returns the PredicatePlugin if one is associated with the provided Address, else undefined
+   */
+  getPlugin(address: string): Promise<PredicatePlugin | undefined>
```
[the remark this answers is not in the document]
willmeister: Same question for error cases in `StateManager.executeTransaction(...)` as well as in general. 

I suppose they all require checks after method invocation, but neither convey information very well unless we start defining our own `Error`s or wrap our return objects in a `Result<T>` that contains custom error code and message or something.

cc: @karlfloersch @ben-chain

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Validates that a transaction has all of the necessary fields populated for it to be useful to StateManager.
+   * Mostly just guards against NPEs.
+   *
+   * @param transaction the Transaction to inspect
+   * @returns true if valid, false otherwise
+   */
+  private validateTransaction(transaction: Transaction): boolean {
```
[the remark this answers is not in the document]
willmeister: Should be addressed. Pulled this out into type-guards.ts. Let me know if you have any other feedback 👍

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Validates that a VerifiedStateUpdate has all of the necessary fields populated for it to be useful to StateManager.
+   * Mostly just guards against NPEs.
+   *
+   * @param verifiedUpdate the VerifiedStateUpdate to inspect
+   * @returns true if valid, false otherwise
+   */
+  private validateVerifiedStateUpdate(verifiedUpdate: VerifiedStateUpdate) : boolean {
```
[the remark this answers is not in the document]
willmeister: Should be addressed. Pulled this out into type-guards.ts. Let me know if you have any other feedback 👍

packages/core/src/app/client/state-db.ts
```diff
@@ -0,0 +1,20 @@
+import BigNum = require('bn.js')
+
+import {StateDB} from "../../interfaces/client";
```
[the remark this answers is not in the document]
willmeister: Added and fixed 👍

packages/core/src/app/client/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './state-db'
+export * from './state-manager'
+export * from './plugin-manager'
```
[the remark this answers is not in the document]
willmeister: Should be addressed now

packages/core/src/app/client/state-manager.ts
willmeister: Was just going off of the [spec](http://spec.plasma.group/en/latest/src/05-client-architecture/state-manager.html#ingesthistoryproof) here. Happy to change it how you all think is best until I'm far enough along to form my own thoughts.

packages/core/src/interfaces/common/utils/range-store.interface.ts
willmeister: As discussed, will make this a larger change in a future PR to merge together `core` and `utils` (tracked in #278)

packages/core/src/app/client/state-manager.ts
karlfloersch: I think long term the efficiency of snapshots is unquestionably a big win. However, think it's important to keep V1 as simple & bug free as possible which means avoiding complex optimizations. That way we can use outputs from the inefficient but (hopefully) bug-free version to test the outputs of our more complex & efficient implementations.

packages/core/src/interfaces/client/plugin-manager.interface.ts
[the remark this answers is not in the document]
karlfloersch: Throwing an error is a bit more expressive & is done in [Level](https://github.com/Level/level#dbgetkey-options-callback) so it's probably a bit better. 

> I suppose they all require checks after method invocation.... start defining our own Errors

Yeah agreed. My guess is defining a custom error is probably the "best practice" but there are many ways to skin a cat.

ben-chain:
Nice, this is coming along!! :D I think we need to change the Transaction interface though--I have been using something different [here](https://github.com/ben-chain/pigi/blob/feat/transition_predicate/packages/contracts/contracts/DataTypes.sol).

We at least need [these parameters](http://spec.plasma.group/en/latest/src/07-predicates/simple-ownership.html#send) for the ownership transaction.  Technically, we also have a `methodId` in the spec right now, but I'm curious if it's not a good idea to just wrap this into a `parameters` field which encompasses all the custom data for that predicate's transaction type.  So for predicates which have multiple transaction methods, we could add a `methodId` as the first parameter but omit it otherwise.  Would be curious of everyones' thoughts on that!

ben-chain:
Was just re-reviewing this morning and caught another small thing--I believe [lines 65/66](https://github.com/willmeister/pigi/blob/feat/235/StateManager-ImplementExecuteTransaction/packages/core/src/app/client/state-manager.ts#L65) should be removed. The reason is that these transactions do not have to be a sub- or super-set of the verified ranges, we just only update their intersection.

For example, If we have only verified `[10, 20)` and get a transaction resulting in some SU on `[0,30)`, then we still update `[10, 20)` as verified into the next block.  Or, if we have verified `[0, 30)` and get a transaction resulting in an SU on `[10, 20)`, we still count `[10, 20)` as verified into the next block.  This is ensured [here](https://github.com/willmeister/pigi/blob/feat/235/StateManager-ImplementExecuteTransaction/packages/core/src/app/client/state-manager.ts#L80).

(Which reminds me--is there an easy way to check that `start < end` any time we instantiate a `Range`?  In this case, because `getVerified...` is pulling only ranges which intersect the transaction, we don't have to worry about it.  But in general this could save some headaches.)

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,105 @@
...
+        // log here?
+        continue
+      }
+
+      const {
+        start: verifiedStart,
+        end: verifiedEnd,
+      }: Range = verifiedUpdate.stateUpdate.id
+      if (
+        transaction.block !== verifiedUpdate.verifiedBlockNumber + 1 ||
```
karlfloersch: The check here should be removed as we no longer require transactions to include a block number.

packages/core/src/app/client/state-manager.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      const predicatePlugin: PredicatePlugin = await this.pluginManager.getPlugin(
+        verifiedUpdate.stateUpdate.newState.predicate
+      )
+
+      const computedState: StateUpdate = await predicatePlugin.executeStateTransition(
+        verifiedUpdate.stateUpdate,
+        transaction
+      )
```
karlfloersch: The verified block number should instead be passed into the `executeStateTransition` & then if we want we can sanity check that the resulting `StateUpdate.blockNumber == verifiedUpdate.verifiedBlockNumber + 1`

willmeister:
> Was just re-reviewing this morning and caught another small thing--I believe [lines 65/66](https://github.com/willmeister/pigi/blob/feat/235/StateManager-ImplementExecuteTransaction/packages/core/src/app/client/state-manager.ts#L65) should be removed. The reason is that these transactions do not have to be a sub- or super-set of the verified ranges, we just only update their intersection.
> 
> For example, If we have only verified `[10, 20)` and get a transaction resulting in some SU on `[0,30)`, then we still update `[10, 20)` as verified into the next block. Or, if we have verified `[0, 30)` and get a transaction resulting in an SU on `[10, 20)`, we still count `[10, 20)` as verified into the next block. This is ensured [here](https://github.com/willmeister/pigi/blob/feat/235/StateManager-ImplementExecuteTransaction/packages/core/src/app/client/state-manager.ts#L80).
> 
> (Which reminds me--is there an easy way to check that `start < end` any time we instantiate a `Range`? In this case, because `getVerified...` is pulling only ranges which intersect the transaction, we don't have to worry about it. But in general this could save some headaches.)

@ben-chain the intention here is to make sure that they do intersect, not that they're sub-/super-sets of each other. If range1.end < range2.start or range1.start > range2.end, there is 0 overlap, right?

`Range` is just a datatype interface, so I'm not sure that's possible at construction time unless we use a factory for it. The alternative is to use [isValidRange](https://github.com/plasma-group/pigi/pull/276/files#diff-8aa9916cfc4b9e19743bb84bb630ba3bR46) or something similar when validation is necessary.

ben-chain:
Ahhhhhh I see--I misread this, I thought it was comparing a start to a start and end to an end.  This is right then, nice!!!

ben-chain:
Amazing!!! :D I think this is getting close to a merge.  Two things:

1) We need an arbitrary `parameters: bytes` in the transactions
2) I think the way to pass the block number into the predicate plugin's transition function is to pass in an arbitrary block number `inBlock` which tells `executeTransaction` which block to execute the transaction in.  This would be the thing checked against `verifiedBlockNumber + 1` and also passed to the predicate plugin.

karlfloersch:
> If range1.end < range2.start or range1.start > range2.end

There's also an `intersects()` function in rangeDB, so it might be good to pull it out and put it in a shared library considering it's such common functionality -- https://github.com/plasma-group/pigi/blob/751d18e8ec2834047e11c0d36cd9af4b338ae1af/packages/operator/src/app/db/range-db.ts#L120-L137

willmeister:
> Amazing!!! :D I think this is getting close to a merge. Two things:
> 
> 1. We need an arbitrary `parameters: bytes` in the transactions
> 2. I think the way to pass the block number into the predicate plugin's transition function is to pass in an arbitrary block number `inBlock` which tells `executeTransaction` which block to execute the transaction in.  This would be the thing checked against `verifiedBlockNumber + 1` and also passed to the predicate plugin.

1. That's [there](https://github.com/plasma-group/pigi/pull/276/files#diff-4a8804da6b0a3615e4ee310e87a23b3cR44), just as a string. I can update it.
2. Can add this, no problem 👍

willmeister:
> > If range1.end < range2.start or range1.start > range2.end
> 
> There's also an `intersects()` function in rangeDB, so it might be good to pull it out and put it in a shared library considering it's such common functionality --
> 
> https://github.com/plasma-group/pigi/blob/751d18e8ec2834047e11c0d36cd9af4b338ae1af/packages/operator/src/app/db/range-db.ts#L120-L137

Adding this to #278 since it falls under better sharing / utils.

PR #277, opened 2019-06-12
Created WalletDB

### LINKED ISSUE #256: Implement WalletDB
[WalletDB](http://spec.plasma.group/en/latest/src/05-client-architecture/wallet-db.html) is a database wrapper used by the [Wallet](http://spec.plasma.group/en/latest/src/05-client-architecture/wallet.html) component of the generalize plasma client. We need to create an interface for `WalletDB` and implement it in the client.

Tasks:
- [ ] Create `WalletDB` interface.
- [ ] Implement `WalletDB` interface.

kfichter:

Fixes #256 

## Description
Implementation of WalletDB + some additional tests. It's pretty simple.


packages/core/test/app/client/wallet-db.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/* External Imports */
+const memdown = require('memdown')
+
+/* Internal Imports */
+import { Keystore } from '../../../src/interfaces'
+import { BaseDB } from '../../../src/app/common/db'
+import { DefaultWalletDB } from '../../../src/app/client/wallet-db'
+
+const keystore: Keystore = {
```
karlfloersch: Have you taken a look at `eth-crypto` specifically [this (encryption)](https://github.com/pubkey/eth-crypto#encryptwithpublickey) and [this (create identity)](https://github.com/pubkey/eth-crypto#createidentity) for keystore creation / encryption? According to https://github.com/ethers-io/ethers.js/pull/233 it's used in ethers.js.

I ask because this config looks pretty low level & I hope there are higher level libraries that should work well enough

karlfloersch: Or I know there are even easier approaches if we're willing to accept the web3 bloat--https://web3py.readthedocs.io/en/stable/web3.eth.account.html#extract-private-key-from-geth-keyfile

packages/core/src/interfaces/client/wallet-db.interface.ts
```diff
@@ -0,0 +1,8 @@
+/* Internal Imports */
+import { Keystore } from '../common'
+
+export interface WalletDB {
+  putKeystore(keystore: Keystore): Promise<void>
+  getKeystore(address: string): Promise<Keystore>
+  listAddresses(): Promise<string[]>
+}
```
karlfloersch: Dope!

packages/core/src/interfaces/common/eth/keystore.interface.ts
```diff
@@ -0,0 +1,76 @@
...
+    kdfparams: {
+      dklen: number
+      n: number
+      p: number
+      r: number
+      salt: string
+    }
+    mac: string
+  }
+}
```
karlfloersch: Can we use a library here?

packages/core/src/app/common/utils/expression.ts
```diff
@@ -0,0 +1,125 @@
+import { Expression } from '../../../interfaces'
```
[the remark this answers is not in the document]
kfichter: Eh yeah I have a messy commit history that I need to fix. It's not reliant on `expressions.ts` but there's some useful cleanup in that commit. Will break up the cleanup and the expressions stuff in a bit.

packages/core/src/interfaces/common/eth/keystore.interface.ts
kfichter: Library for what part? Generally trying to reduce dependence on libraries when possible.

packages/core/test/app/client/wallet-db.spec.ts
kfichter: Yeah I wouldn't create this manually - just did it here for the sake of testing. `Wallet` (not implemented in this PR) will use a library (probably eth-crypto) to generate keystore files.

packages/core/src/interfaces/common/eth/keystore.interface.ts
karlfloersch: Ah I just figured this could use a library like `eth-crypto`

kfichter: Ahh ok. I looked at both `eth-crypto` and `ethers` just now but couldn't find an interface for `Keystore` so I think it's fine to just keep it here.

packages/core/src/app/common/utils/expression.ts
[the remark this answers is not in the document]
kfichter: Fixed

packages/core/src/app/client/wallet-db.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Adds a keystore file to the database.
+   * @param keystore Keystore file to add.
+   * @returns a Promise that resolves once the keystore has been inserted.
+   */
+  public async putKeystore(keystore: Keystore): Promise<void> {
+    const key = Buffer.from(keystore.address)
+    const value = Buffer.from(JSON.stringify(keystore))
+
+    if (await this.hasKeystore(keystore.address)) {
```
willmeister: Don't know the context in which it's used, but I think there may be a race condition here.

kfichter: I think it should be fine this instance since there's (effectively) a 1:1 mapping between addresses and keystore files.

kfichter: Honestly could probably just remove the error check, no real reason to have it there.

kfichter:
@karlfloersch @willmeister does this look ready to merge?

PR #280, opened 2019-06-13
[WIP] Added initial BlockManager and BlockDB implementations

### LINKED ISSUE #266: Implement BlockDB
[BlockDB](http://spec.plasma.group/en/latest/src/06-operator-architecture/block-db.html) is a database wrapper used by the [BlockManager](http://spec.plasma.group/en/latest/src/06-operator-architecture/block-manager.html) to store information about the next block to be published.

Tasks:
- [ ] Create an interface for `BlockDB`.
- [ ] Implement `BlockDB`.

### LINKED ISSUE #267: Implement BlockManager
[BlockManager](http://spec.plasma.group/en/latest/src/06-operator-architecture/block-manager.html) is a service that regularly creates new blocks and publishes block headers to Ethereum. Block manager should check for transactions every X seconds (client defined). If there are new transactions, the block manager will compute the Merkle root for the next block and publish the root to a commitment contract.

Tasks:
- [ ] Create an interface for `BlockManager`.
- [ ] Implement `BlockManager`.

### LINKED ISSUE #268: Implement Merkle Interval Tree
The [Merkle Interval Tree](http://spec.plasma.group/en/latest/src/01-core/merkle-interval-tree.html) is a special type of Merkle tree that allows users to create proofs of inclusion/exclusion for *ranges* of items. We need to implement the Merkle Interval Tree according to the [interface we've already specified](http://spec.plasma.group/en/latest/src/05-client-architecture/merkle-interval-tree.html).

Tasks:
- [ ] Create an interface for `MerkleIntervalTree`.
- [ ] Implement `MerkleIntervalTree`.

kfichter:
## Description
Started implementing [BlockManager](http://spec.plasma.group/en/latest/src/06-operator-architecture/block-manager.html) and [BlockDB](http://spec.plasma.group/en/latest/src/06-operator-architecture/block-db.htm). Still **WIP** but want to get review as I go along. Partially blocked by #271 at the moment because `RangeStore` is currently located in `Operator` but I don't want to do an import when we're moving it soon anyway.

## Questions
- [ ] How do we want to handle resource locking on the database? For example, `BlockDB` needs to lock a specific range so that a race condition doesn't cause a database overwrite. Can this be handled at the `LevelDB` level? How important is this?
- [ ] Should `BlockManager` maintain an append-only log like it did back in the original operator?
- [ ] Should `BlockManager` be writing to `StateManager` or `HistoryManager` once it publishes a block? It doesn't really make sense for both `BlockManager` and `HistoryManager` to store the full set of state updates (?). 

### Fixes
- Fixes #266
- Fixes #267
- Fixes #268

### Blockers
- [ ] #268 
- [ ] #271 

## Contributing Agreement


packages/core/src/app/operator/block-db.ts
```diff
@@ -0,0 +1,146 @@
...
+   */
+  public async addPendingStateUpdate(stateUpdate: StateUpdate): Promise<void> {
+    const block = await this.getNextBlockStore()
+    const start = stateUpdate.id.start
+    const end = stateUpdate.id.end
+
+    // TODO: Figure out how to implement locking here so two state updates
+    // can't be added at the same time.
+
+    if (await block.has(start, end)) {
```
willmeister: How frequently do we imagine Operators will submit plasma blocks? There's still a lot for me to learn, but this line triggers some red flags around scalability, depending on the answer.

packages/core/src/app/operator/block-db.ts
```diff
@@ -0,0 +1,146 @@
...
+      leaves,
+    })
+    return tree.root
+  }
+
+  /**
+   * Finalizes the next plasma block so that it can be published.
+   * @returns a promise that resolves once the block has been finalized.
+   */
+  public async finalizeNextBlock(): Promise<void> {
```
willmeister: Not sure that it matters, but there's a race condition here if called concurrently.

packages/core/src/app/operator/block-db.ts
kfichter: Probably somewhere between 15sec and 10mins depending on the chain. Which part specifically raises red flags?

packages/core/src/app/operator/block-db.ts
kfichter: Yeah I could definitely use some input on managing concurrency. Seems like we might have to manage it on a global/OS level because `leveldb` isn't safe between different processes.

kfichter: Specifically important for the operator because accidentally bumping the block number twice would break a lot of things.

packages/core/src/app/operator/block-db.ts
willmeister: Sorry @kfichter, just seeing this now, but it means that the same range can only be transacted once every 10-15sec. That's probably fine for now, but I can certainly think of applications for which that will be too slow (trading, poker, etc.).

packages/core/src/app/operator/block-db.ts
willmeister: We'll have to look into this more. It could be as simple as using a lib that wraps some simple single-threaded bool-flipping logic. We should be able to use JS's single-threaded nature to enforce order where necessary on multiple awaited calls.

PR #281, opened 2019-06-14
Consolidating utils into core package and updating references

### LINKED ISSUE #278: Merge together utils and core packages where applicable
Right now `core` and `utils` should both only be depended on and have no dependencies. But they sort of depend on each other and/or serve the same function in some places.

This task is to merge the two where possible, and most importantly make them have different purposes. If that means one absorbs the other, great, but the best solution may not require that.

See [this](https://github.com/plasma-group/pigi/pull/276#discussion_r293062094) for more context

willmeister:
## Description
This PR consolidates the `utils` package into the `core` package and updating all necessary references.

Note: To make this easy, I had to slightly modify the `Transaction` datatype, but that will get redefined by existing PRs, which is why I didn't want to bother with discrepancies now.

## Questions
- When should we merge this in so as to not break all of the things and existing PRs?

### Fixes
- Fixes #278

## Contributing Agreement


kfichter:
LGTM

PR #282, opened 2019-06-15
Feat/state manager fixes

ben-chain:
## Description
Very small PR for the state manager to pass `inBlock` to the predicate plugin instead of the `verifiedBlockNumber`.

## Contributing Agreement


kfichter:
I think this PR needs to be rebased onto the latest commit, otherwise Travis will fail.

PR #283, opened 2019-06-16
Added MerkleIntervalTree implementation

### LINKED ISSUE #268: Implement Merkle Interval Tree

kfichter:
## Description
Implements [`MerkleIntervalTree`](http://spec.plasma.group/en/latest/src/01-core/merkle-interval-tree.html) as described in our specification. Required some slight tweaks to the `keccak256` function (`Buffer => Buffer` instead of `string => string`). Added tests as well.

### Fixes
- Fixes #268 

## Contributing Agreement


ben-chain:
Woot woot!  A couple notes:
- the base class probably shouldn't parse leaves as ``start, end, data`` but instead ``start, data``. The state tree does have this special ``end`` but the asset tree does not.  So maybe it's best to add that in some ``stateTree`` child class, where it can parse from the SU and add the salt too.
-  it might be worth replacing the read/writes to ``levels[][]`` with something like ``get/setNodeValueAt(uint,uint)`` which wraps the array stuff.  That way, it would be very simple to extend for DB-based storage.

kfichter:
> * the base class probably shouldn't parse leaves as `start, end, data` but instead `start, data`. The state tree does have this special `end` but the asset tree does not.  So maybe it's best to add that in some `stateTree` child class, where it can parse from the SU and add the salt too.

Hmmm my understanding was that in the asset tree `start == end` and that's why there's no special end value. What if `end` is optional and defaults to `start` if not provided (since that's what's being implied anyway)?

Also `salt` could simply be added to `data` before creating the tree so I don't think it's necessary to actually explicitly have a salt parameter.

> * it might be worth replacing the read/writes to `levels[][]` with something like `get/setNodeValueAt(uint,uint)` which wraps the array stuff.  That way, it would be very simple to extend for DB-based storage.

Yeah this is a good point, will do.

packages/core/src/app/common/utils/merkle-interval-tree.ts
```diff
@@ -0,0 +1,332 @@
...
+        rightChild = computedNode
+      } else {
+        leftChild = computedNode
+        rightChild = siblingNode
+
+        if (
+          (prevRightSibling !== null &&
+            rightChild.index.lt(prevRightSibling.index)) ||
+          rightChild.index.lt(leafNode.end)
+        ) {
```
karlfloersch: Does this cover every case? Or just a bunch of early error detection?

packages/core/src/app/common/utils/merkle-interval-tree.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    const firstRightSiblingIndex = path.indexOf('0')
+    const firstRightSibling =
+      firstRightSiblingIndex >= 0
+        ? inclusionProof[firstRightSiblingIndex]
+        : null
+
+    const implicitStart = leafPosition === 0 ? new BigNumber(0) : leafNode.start
+    const implicitEnd =
+      firstRightSibling !== null ? firstRightSibling.index : null
```
karlfloersch: Maybe a comment on what "implicit" means here

packages/core/src/app/common/utils/merkle-interval-tree.ts
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  /**
+   * Gets the root and implicit bounds for a given leaf.
+   * @param leafNode Leaf to get root and bounds for.
+   * @param leafPosition Index of the leaf in the list of leaves.
+   * @param inclusionProof Inclusion proof for the leaf.
+   * @returns the root and bounds for the leaf.
+   */
+  public getRootAndBounds(
```
karlfloersch: Generally a bunch of comments here would be great

packages/core/test/app/common/utils/merkle-interval-tree.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+      const root = tree.getRoot()
+
+      root.should.deep.equal({
+        index: new BigNumber(0),
+        hash: hexify(
+          'e1b53cab461af771ad8d060145d2e27a04ee7c2e671efe4feac748de8cef1fc5'
+        ),
+      })
+    })
+  })
```
karlfloersch: Another test could check that an error is thrown if leaf nodes have invalid ranges (`start > end`)

packages/core/test/app/common/utils/merkle-interval-tree.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+          end: new BigNumber(200),
+          data: hexify('5678'),
+        },
+      ])
+
+      should.Throw(() => {
+        tree.getInclusionProof(2)
+      }, 'Leaf position is out of bounds.')
+    })
+  })
```
karlfloersch: Once again could verify that an error is thrown when `start > end` for any of these ranges

kfichter: Hmmm I think unnecessary here but added in `checkInclusionProof`.

packages/core/src/app/common/utils/merkle-interval-tree.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Checks if two ranges overlap.
+ * @param rangeA First range to check.
+ * @param rangeB Second range to check.
+ * @returns `true` if the ranges overlap, `false` otherwise.
+ */
+const intersects = (rangeA: Range, rangeB: Range): boolean => {
+  const maxStart = bnMax(rangeA.start, rangeB.start)
+  const minEnd = bnMin(rangeA.end, rangeB.end)
+  return maxStart.lt(minEnd)
+}
```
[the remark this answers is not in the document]
kfichter: Hmm for now just going to use the one in `range.ts` then and we can change later.

karlfloersch:
Actually after discussions would be best to add a salt in the block prod / verify inclusion. The reasoning is that if you put the salt inside the state object then we need to verify signatures over more data which could be non-trivial overhead at large scale.

PR #286, opened 2019-06-18
Implemented Wallet

### LINKED ISSUE #260: Implement Wallet
[Wallet](http://spec.plasma.group/en/latest/src/05-client-architecture/wallet.html) is a simple component that allows the user to create and manage private keys. We need to create a TypeScript interface for `Wallet` and implement it.

We also need to make a few changes to the `Wallet` API so that it can support locking/unlocking accounts. Note that `Wallet.createAccount` **MUST** encrypt the user's account.

Tasks:
- [ ] Update `Wallet` API on spec.
- [ ] Create an interface for `Wallet`.
- [ ] Implement `Wallet`.

kfichter:
## Description
[Wallet](http://spec.plasma.group/en/latest/src/05-client-architecture/wallet.html) is a service that allows users to create and manage Ethereum accounts. This PR implements `Wallet` according to the provided specification. Requires a few modifications to `WalletDB`.

### Fixes
- Fixes #260 

### Blockers
- #277 

## Contributing Agreement


packages/core/test/app/client/wallet.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+/* Internal Imports */
+import { Wallet } from '../../../src/interfaces'
+import { DefaultWallet, DefaultWalletDB, BaseDB } from '../../../src/app'
+
+describe('DefaultWallet', () => {
+  let walletdb: DefaultWalletDB
+  let wallet: DefaultWallet
+
+  beforeEach(() => {
+    walletdb = new DefaultWalletDB(new BaseDB(new MemDown('') as any))
```
karlfloersch: What does the `as any` do here? Asking out of curiosity.

packages/core/test/app/client/wallet.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      await wallet.lockAccount(account).should.be.fulfilled
+    }).timeout(5000)
+
+    it('should lock an account even if not unlocked', async () => {
+      const account = await wallet.createAccount('password')
+
+      await wallet.lockAccount(account).should.be.fulfilled
+    }).timeout(5000)
+  })
+})
```
karlfloersch: For completeness could add a test for `sign`

PR #299, opened 2019-06-23
Clean up structure & remove orphaned code

### LINKED ISSUE #298: Remove orphaned code / refactor core
We've got a lot of dead/orphaned code & core currently has a bunch disorganized directories. This should change!

Remove dead code & restructure core.

karlfloersch:
## Description
This PR includes an extensive refactor of pigi. It includes:

- Deleted unused packages
  - Removed vyper-js as we are no longer using vyper to write our smart contracts
  - Removed plasma-js as it was mostly just a skeleton of a library
- Removed sections of core
  - Removed the app structural code located in https://github.com/plasma-group/pigi/tree/master/packages/core/src/app/common/app and https://github.com/plasma-group/pigi/tree/master/packages/core/src/app/core/app as it was unused & untested (we don't have integration tests or a single main app). Something like this should exist eventually but until that point it would remain orphaned.
- Flattened core's directory structure. The app directory ( https://github.com/karlfloersch/pigi/tree/clean/remove_orphaned_code/packages/core/src/app  ) now contains many of the directories which were put under `core` and `common`. Less looking through directories yay!
  - Notice that there is a `block-production` directory and an `ovm` directory. This is *as opposed* to the past naming scheme of `state-manager` and `block-manager`. I must say that there is a symmetry to the old names, but I also do feel that these are more descriptive directory names. However, I am very interested what you think @ben-chain . 
- Renamed `interfaces` to `types`
  - The issue with calling it interfaces is I wanted to group 3 things together: 1) interfaces, 2) type definitions, and 3) type guards. Note that this includes *type guards* which was previously put in `utils`. I am not entirely convinced that putting type guards in the types directory is right, but it felt good.
  - Flattened the types (formerly interfaces) directory. This means breaking the convention that we'd have a mirrored directory structure for types. It was becoming a hassle to shuffle through all these interface directories. Instead just grouping types by files or directories, up to developer discretion is a reasonable thing to do. Would love feedback here as well!
- Broke up this absolutely terrifying utils directory 😅 https://github.com/plasma-group/pigi/tree/master/packages/core/src/app/common/utils
- Added tests to the client/server code
  - Also created a class `SimpleClient` and `SimpleServer` which have sensible defaults for a REST-based json rpc client & server.

I'd love feedback on:
- naming of the core root directories
  - state/block manager vs ovm & block production
- where to put typeguards
- where to put interfaces even
- naming the client / server code `transport` vs something like `network` or `client-server` or ??

### Fixes
- Fixes #298 

## Contributing Agreement


karlfloersch:
> would we say "the state and history managers are a part of the OBM"? That makes sense to me but wanna see if this is what you mean.

Yeah agreed -- state manager & history manager live in the OVM.

PR #304, opened 2019-06-25
Removing it.only from merkle-interval-tree.spec.ts so all core tests run

willmeister:
## Description
Fixing a bug where not all Core tests are run

## Contributing Agreement


packages/core/test/app/block-production/merkle-interval-tree.spec.ts
```diff
@@ -158,7 +158,7 @@ describe('merkle-index-tree', () => {
       const plasmaBlock = new PlasmaBlock(blockContents)
       log(plasmaBlock)
     })
-    it.only('should generate and verify a StateUpdateInclusionProof', async () => {
```
karlfloersch: Uh oh! Sorry that snuck past me in review -- great catch and merging now.

PR #312, opened 2019-06-26
Implement Ownership Predicate Plugin (292)

### LINKED ISSUE #292: Implement & test ownership predicate plugin state transition
Add to the state manager a simple predicate plugin for ownership. The most critical function being `executeStateTransition()`. Once this is implemented it is possible to more throughly test the state manager by applying ownership transactions which allows us to test actually sending funds from one account to another.

For now we will not implement the query language. Instead, only for testing, we can instead just add a function which returns all ranges "owned" by a particular user. This function is simply used for debugging & not expected to be performed once the aggregator is live, and likely just accepts a RangeBucket.

An outline of the predicate plugin format can be found here: http://spec.plasma.group/en/latest/src/05-client-architecture/predicate-plugin.html

willmeister:
## Description
Adding the OwnershipPredicatePlugin class that handles the logic for simple ownership of a range of coins. 

## Questions
- Block numbers should be BigNums, right?
- Any utils currently exist for pub key crypto?

### Fixes
- Fixes #292

## Contributing Agreement


ben-chain:
Haven't looked over the code yet, but addressing the two questions:
1) I'm not sure we will ever produce enough blocks to exceed the JS ``number``, but it's probably best practice.
2) I don't think we have any signing utils at the moment--I am guilty of not yet checking sigs in the ownership contract 😅

packages/predicates/src/plugins/ownership-predicate.ts
```diff
@@ -0,0 +1,104 @@
...
+    transaction: Transaction,
+    witness: string
+  ): Promise<StateUpdate> {
+    await this.validateStateTransition(
+      previousStateUpdate,
+      transaction,
+      witness
+    )
+
+    const range = getOverlappingRange(
```
akhila-raju: Isn't range supposed to be transaction.start to transaction.end? See [docs](http://spec.plasma.group/en/latest/src/07-predicates/simple-ownership.html#state-transitions)

willmeister: Good call -- I was just worried about checking that they overlap. Also need to make sure that the `transaction.range` is a subset of the `previousStateUpdate.range`

akhila-raju:
LGTM but @ben-chain or @karlfloersch should approve!

packages/predicates/src/plugins/ownership-predicate.ts
```diff
@@ -0,0 +1,101 @@
...
+          transaction
+        )}] because transaction range is not a subset of previous state update range.`
+      )
+    }
+
+    const lastSyncedBlock: BigNum = await this.syncManager.getLastSyncedBlock(
+      previousStateUpdate.stateObject.predicateAddress
+    )
+
+    if (lastSyncedBlock.gte(transaction.parameters.targetBlock)) {
```
ben-chain: Ah shoot--we've just run into an issue where my irresponsible PR didn't update the docs to reflect changes in the contract--my bad!  We discovered in implementing the contract that ``targetBlock`` isn't needed because the deprecation call never actually checks where the block is included.  However, this still is worth commenting on, since if we did need ``targetBlock``, it would have to be done a little differently.  

The sync manager is meant to be a coordinator for the state manager, but calls within the state manager itself should not use it--for example, if we had previously synced block 100, and then received a new coin, we would need to use this function to verify history in blocks 1-99 and not care about having already synced to a different block.  In fact, with the new model for message queue/PubNub stuff discussed with Matt yesterday, we may not need the sync manager at all, and instead the message queue will automatically tell us exactly what to "sync" based on any incoming messages which involve us e.g. receiving new ranges.

I know that's not quite the check you're doing here--if it's greater than the last synced block, clearly we're skipping history and doing something wrong.  Luckily, that check is automatically handled by the ``verifiedBlockNumber`` check in ``executeTransaction`` itself.  Basically, since we end up verifying different block heights for different ranges, any notion of a globally synced block is irrelevant for a particular transaction verification--we *always* need to check that we have verified up to the last block before accepting a transaction, and so this is handled always via ``verifiedStateUpdates`` in the block manager, and we can ignore it for predicates and just deal with the transition logic. :)

packages/predicates/src/plugins/ownership-predicate.ts
```diff
@@ -0,0 +1,101 @@
...
+      previousStateUpdate,
+      transaction,
+      witness
+    )
+
+    return {
+      range: transaction.range,
+      stateObject: transaction.parameters.newState,
+      depositAddress: transaction.depositAddress,
+      plasmaBlockNumber: transaction.parameters.originBlock,
```
ben-chain: The ``plasmaBlockNumber`` for the output of the state transition should always be the``inBlock``--because "that is the block we are executing the transaction in."  

For color, the ``originBlock`` is for the safety check run below, to prevent transaction "replay"--if Alice signs a send to Bob from ``targetBlock==0`` and ``inBlock==1``, and then Bob sends the  coin back to her with ``targetBlock==1`` and ``inBlock==2``, the check guarantees that her previous transaction could not be re-applied because the only valid sends from that new state update would have to have ``originBlock >= 2``.

It's a little confusing--realistically, literally every single predicate plugin will end up returning a state update with ``plasmaBlockNumber === inBlock``.  I thought in the past about having some sort of type which is a state update without the block, but that seemed weird too.  Maybe in reality the predicate plugin only needs to return a ``StateObject``?  And then the state manager would reconstruct the state update based on ``inBlock``, ``transaction.depositAddress``, ``transaction.range``, and the ``stateObject`` output by the ``executeStateTransition``.  Hmmmm... that would definitely work here for ownership, but I'm not sure if it's future-proof.  Paging @karlfloersch for thoughts on how this would work with swaps and whether just returning a state object is sufficient.

willmeister: I like that idea a lot to reduce duplication. Is there ever a case where range, block, or depositAddress would change? I can't think of any, but I could be missing something.

willmeister: Going to leave it as returning a `StateUpdate` for now, but happy to change it if we reach consensus.

ben-chain: Sorry, been in transit to Akhila--just confirmed with Karl, he also thinks that using a state object is cleaner and fine for now 👍

akhila-raju: Since `StateObject` is the only field being mutilated, makes sense for `executeStateTransition` to output a `StateObject` rather than a `StateUpdate`. Thanks for the explanation @ben-chain

packages/predicates/src/plugins/ownership-predicate.ts
```diff
@@ -0,0 +1,93 @@
...
+      throw new Error(
+        `Cannot transition from state [${JSON.stringify(
+          previousStateUpdate
+        )}] with transaction [${JSON.stringify(
+          transaction
+        )}] because block number [${transaction.body.originBlock.toNumber()}] is not greater than previous state block number.`
+      )
+    }
+
+    if (!isRangeSubset(transaction.range, previousStateUpdate.range)) {
```
ben-chain: Counterintuitively, we actually need to remove this check.  Any problems related to the transaction being larger than the SU will be handled by the state manager.  The reason we need to get rid of it is it allows us to merge ranges: If Alice receives ``[0,5)`` in block 1, and ``[5,10)`` in block 2, then she is permitted to make a transaction on ``[0,10)`` with ``originBlock = 2``.

packages/predicates/src/plugins/ownership-predicate.ts
willmeister: This is updated now 👍

packages/predicates/src/plugins/ownership-predicate.ts
willmeister: Yeah, that makes perfect sense. Thanks! 👍

willmeister:
> Haven't looked over the code yet, but addressing the two questions:
> 
> 1. I'm not sure we will ever produce enough blocks to exceed the JS `number`, but it's probably best practice.
> 2. I don't think we have any signing utils at the moment--I am guilty of not yet checking sigs in the ownership contract 😅

@ben-chain re: 2, can we agree to a method signature so that we can create a dummy function for it and reference it everywhere so we don't have to update a ton of places when we actually create it?Maybe: 
```
verify(signature: string, message: string, pubKey: string): boolean
```
?

ben-chain:
Yes, sounds great!  In the long run we probably need a proper ``Signature`` type with (v,r,s) and whatnot, but could even be left as ``any`` for now.

PR #319, opened 2019-07-06
Add range bucket iterator

### LINKED ISSUE #318: Add iterator() to RangeBucket
RangeBucket currently has not implemented `iterator()` which could be useful.

Calling iterator we return an iterator which handles the conversion into RangeEntries automatically (it doesn't iterate over raw key/values as they must be processed by RangeDB to be more human readable).

karlfloersch:
## Description
This adds an `iterator()` function to the RangeBucket. This iterator is a bit custom because it has a function `nextRange()` which instead of returning a raw KV result, it instead parses it and returns a `RangeEntry`.

This feature is particularly useful if we ever want to iterate over the full contents of a RangeDB and return proper `RangeEntries`. I needed this for the way I was going to build the merkle tree creator using LevelDB. However, realized that our blocks won't get large enough to need to use a DB like this and so have abandoned that. Still this could be a useful feature!

## Questions
- Yall know of any cleaner ways to return this iterator other than extending our BaseIterator class?
- Should we make `next()` return a Range by default instead of calling it `nextRange()`? And maybe add `nextRaw()` which acts like the normal next function?

### Fixes
- Fixes #318

## Contributing Agreement

### DIFF
--- a/packages/core/src/app/db/range-bucket.ts
+++ b/packages/core/src/app/db/range-bucket.ts
@@ -15,15 +15,17 @@ import {
   Iterator,
   K,
   V,
+  KV,
   RangeBucket,
   RangeEntry,
   Endianness,
 } from '../../types'
-import { bufferUtils, intersects, BaseDB } from '../../app'
+import { bufferUtils, intersects, BaseDB, RangeIterator } from '../../app'
 
 /* Logging */
 import debug from 'debug'
 const log = debug('info:range-db')
+
 
 /**
  * Simple bucket implementation that forwards all
@@ -133,7 +135,7 @@ export class BaseRangeBucket implements RangeBucket {
    * @param result The resulting value which has been extracted from our DB.
    * @returns a range object with {start, end, value}
    */
-  private resultToRange(result): RangeEntry {
+  private resultToRange(result: KV): RangeEntry {
     // Helper function which gets the start and end position from a DB seek result
     return {
       start: new BigNum(this.getStartFromValue(result.value)),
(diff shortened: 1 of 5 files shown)


packages/core/src/app/db/range-bucket.ts
```diff
@@ -282,6 +284,23 @@ export class BaseRangeBucket implements RangeBucket {
...
+   * Creates an iterator with some options.
+   * @param options Parameters for the iterator.
+   * @returns the iterator instance.
+   */
+  public iterator(options?: IteratorOptions): RangeIterator {
+    const rangeIterator = new BaseRangeIterator(
+      this.db,
+      {
+        ...options,
+        prefix: this.addPrefix(this.prefix),
```
willmeister: Should this just be `this.prefix`, or do we want the prefix twice?

packages/core/test/app/db/range-bucket.spec.ts
```diff
@@ -45,6 +45,22 @@ const testPutResults = async (
   db: RangeBucket,
   putContents: any[],
   expectedResults: any[]
+): Promise<void> => {
+  // First put the ranges
+  putRanges(db, putContents)
+  // Now check that they were added correctly
+  const res = await db.get(
+    new BigNum('0', 'hex'),
```
willmeister: Can use `ZERO`

packages/core/test/app/db/range-bucket.spec.ts
```diff
@@ -256,4 +269,32 @@ describe('RangeDB', () => {
       ]
     )
   })
+
+  describe('iterator()', () => {
+    it('allows nextRange() to be called by the iterator returning a RangeEntry instead of a KV', async () => {
+      const testRanges = {
```
willmeister: I might suggest having the inputs be the expected results. If `db.put(...)` doesn't properly  overwrite range 50-150, this test will fail, even though the `iterator()` call works properly.

packages/core/src/app/db/range-bucket.ts
karlfloersch: Ah good catch! This should be `options.prefix` not `this.prefix` as it is here -- https://github.com/plasma-group/pigi/blob/master/packages/core/src/app/db/bucket.ts#L80-L85

packages/core/test/app/db/range-bucket.spec.ts
karlfloersch: True, though there are a lot of `'0'`s which I'd have to change in this file. Do you think it's worth it to define a special `ZERO` constant?

PR #320, opened 2019-07-07
Add inclusion proof checking to commitment contract

### LINKED ISSUE #263: Create an interface for commitment contracts
[Commitment contracts](http://spec.plasma.group/en/latest/src/02-contracts/commitment-contract.html) hold the list of block headers for a specific plasma chain. The Generalized Plasma Specification describes an interface for commitment contracts, so we should create an [interface](https://solidity.readthedocs.io/en/v0.4.24/contracts.html#interfaces) based on the specified API.

ben-chain:
## Description
This PR adds support for actually submitting blocks to the commitment contract, and verifying inclusion of state updates.

Still need to make it so that the Deposit Contract tests use this function, I left the verify function as still returning `true` for now to not muddle this PR.

## Questions
- What do we think of the naming convention I've been using? (`assetTree`, `stateSubtree`)
- Given the new understanding of the OVM architecture, should this technically be a `verifier` contract separate from the commitment contract?
- Does test coverage seem up to par?

### Fixes
- Fixes #263 

### Blockers
N/A

## Contributing Agreement


packages/contracts/contracts/CommitmentChain.sol
```diff
@@ -5,39 +5,155 @@ pragma experimental ABIEncoderV2;
...
+        // calculate what the block root would be if the stateSubtreeRoot is correct
+        bytes32 assetTreeRoot = verifyAssetTreeInclusionAndGetRoot(stateSubtreeRoot, _stateUpdate.depositAddress, _inclusionProof.assetTreeInclusionProof);
+        // check this is indeed the root for the state update's block
+        uint128 expectedPlasmaBlockNumber = _stateUpdate.plasmaBlockNumber;
+        return assetTreeRoot == blocks[expectedPlasmaBlockNumber];
+    }
+
+    // Via https://github.com/ethereum/solidity-examples/blob/master/src/bits/Bits.sol
+    // Gets the value of the 'index'th bit (where 0 => rightmost bit) in the binary expression of 'self'.
+    function getNthBitFromRight(uint self, uint8 index) public pure returns (uint8) {
```
willmeister: Looks like the result of this is used as a `bool` everywhere -- should the name change to `isNthBitFromRightSet` and the body change to `return (self >> index & 1) == 1`?

packages/contracts/contracts/CommitmentChain.sol
```diff
@@ -5,39 +5,155 @@ pragma experimental ABIEncoderV2;
...
+        uint128 previousRightLowerBound = _stateUpdate.range.end;
+        for (uint8 level = 0; level < _proof.siblings.length; level++) {
+            dt.StateSubtreeNode memory siblingNode = _proof.siblings[level];
+            // the binaryPath up the tree is the leafPosition expressed in bits
+            uint8 isComputedRightSibling = getNthBitFromRight(leafPosition, level);
+            if (isComputedRightSibling == 0) {
+                // if binaryPath[level] == 0, the proof element is a right sibling
+                computedNode = stateSubtreeParent(computedNode, siblingNode);
+                // make sure the sibling does not violate interval proof lowerBound rules
+                require(previousRightLowerBound < siblingNode.lowerBound, 'No valid branch allows potential intersections with other branches.');
```
willmeister: I think this check is only useful the first time it is executed because `previousRightLowerBound` will actually be the _end_ of the left sibling's range. After that, it's set to `siblingNode.lowerBound` which should be the _start_ of the range, making this check [redundant](https://github.com/plasma-group/pigi/pull/320/files#diff-5a51d082885f7952ed4858e1be679f3dR71), right?

karlfloersch:
im sry

PR #325, opened 2019-07-09
Mock ingest transactions & produce blocks

### LINKED ISSUE #323: Mock ingest transactions & produce blocks
The aggregator must accept transactions, and produce blocks with state updates which result from those transactions. This process is independent from the authentication of transactions or the state transition calculations. Therefore we can create an initial version which simply accepts transactions with `ingestTransaction()` where the transaction is simply the StateUpdate which will be included in the next block. These StateUpdates will be added to a queue which is then emptied periodically to produce a next block.

The function calls would look something like:
- `generateNSequentialStateUpdates()` https://github.com/plasma-group/pigi/blob/master/packages/core/test/app/block-production/merkle-interval-tree.spec.ts#L19-L38 on the client. Then send those state updates to...
- `ingestTransaction(transaction)` on the aggregator (right now transactions are just state updates). This would mock `executeTransaction()` and simply push the SU to...
- `enqueueStateUpdate(StateUpdate)` which adds our StateUpdate to a RangeBucket (eg. https://github.com/plasma-group/pigi/blob/master/packages/core/test/app/db/range-bucket.spec.ts#L160 )
- `produceBlock()` which is triggered periodically that first checks if there are stateupdates in the queue (our rangeDB bucket), and then produces a block if this is true.

 Some of this functionality was started in -- https://github.com/plasma-group/pigi/pull/280/files .

willmeister:
## Description
Adds Aggregator and BlockManager (copied from [here](https://github.com/plasma-group/pigi/pull/280/files#diff-94a4ac125235f8f40580dd8fac091308)) and implements and tests `Aggregator.ingestTransaction(...)`

## Questions
- I assume this is supposed to lean heavily on BlockManager for addPendingStateUpdate and submitNextBlock -- is that right, or is there a separate `enqueueStateUpdate(...)` and `produceBlock()` that are required?
- Is there a good way to get around passing the `witness` to `Aggregator.ingestTransaction(...)`?

### Fixes
- Fixes #323

## Contributing Agreement


karlfloersch:
This PR is looking great so far! Glad you pulled out `generateNSequentialStateUpdates()` into test utils--was meaning to do that. With regard to your questions--

> supposed to lean heavily on BlockManager for addPendingStateUpdate and submitNextBlock

Yes this is right. `enqueueStateUpdate()` & `produceBlock()` are just alternative names for `addPendingStateUpdate` & `submitNextBlock` -- Apologies for my imprecise language in the issue!

> Is there a good way to get around passing the witness to `Aggregator.ingestTransaction(...)`?

So at this moment I am not entirely sure if we need the witness at all. @ben-chain let me know what you think, but it seems to me you should only need to send the aggregator a `transaction`, and then if someone wants to validate an OVM property, they'd simply be passed whatever witness they need for that in the `proof`. Plus I'm a fan of removing unused fields where possible & I can't think of a use now so feels like a good thing to get rid of.

Is this what you were asking with this question @willmeister  ?

willmeister:
> ...
> So at this moment I am not entirely sure if we need the witness at all. @ben-chain let me know what you think, but it seems to me you should only need to send the aggregator a `transaction`, and then if someone wants to validate an OVM property, they'd simply be passed whatever witness they need for that in the `proof`. Plus I'm a fan of removing unused fields where possible & I can't think of a use now so feels like a good thing to get rid of.
> 
> Is this what you were asking with this question @willmeister ?

@karlfloersch the reason I added it is that we call `StateManager.executeTransaction(transaction, block, witness)` in order to get the `StateUpdate` [and maybe the valid Ranges?] from the `Transaction`. If this is not necessary and there a better way to get the `StateUpdate`, I'll happily switch to that 😄

ben-chain:
I think that this is the right model, @karlfloersch we will probably end up calling transaction `property` and could rename witness to `proof` if we like.  Naming aside, the function call given our current names is `executeTransaction`

packages/core/src/app/ovm/state-manager.ts
```diff
@@ -24,13 +24,10 @@ import { getOverlappingRange, ONE, rangesIntersect } from '../../app'
...
-  private pluginManager: PluginManager
-
-  public constructor(stateDB: StateDB, pluginManager: PluginManager) {
-    this.stateDB = stateDB
-    this.pluginManager = pluginManager
-  }
+  public constructor(
+    private stateDB: StateDB,
+    private pluginManager: PluginManager
+  ) {}
```
akhila-raju: Does this automatically set these params to private variables within this class?

packages/predicates/test/ownership-predicate.spec.ts
```diff
@@ -126,7 +123,7 @@ describe('OwnershipPredicate', async () => {
       const stateObject: StateObject = await ownershipPredicate.executeStateTransition(
         defaultPreviousStateUpdate,
         defaultTransaction,
-        defaultWitness
+        defaultTransaction
```
akhila-raju: Why is the witness being replaced by a transaction here and below?

packages/core/src/app/utils/range.ts
```diff
@@ -65,6 +65,39 @@ export const isRangeSubset = (subset: Range, superset: Range): boolean => {
   )
 }
 
+/**
+ * Determines whether the provided Ranges collectively span the Range in question
+ *
+ * @param ranges the Ranges that will/won't span the rangeToSpan
+ * @param rangeToSpan the Range being spanned
```
akhila-raju: This param name is a bit confusing, perhaps consider renaming to currentRange?

packages/core/src/app/utils/range.ts
```diff
@@ -65,6 +65,39 @@ export const isRangeSubset = (subset: Range, superset: Range): boolean => {
...
+ * @returns true if ranges span rangeToSpan, false otherwise
+ */
+export const rangesSpanRange = (
```
akhila-raju: I would give the function name here a bit more of a descriptive name

packages/core/src/app/utils/range.ts
```diff
@@ -65,6 +65,39 @@ export const isRangeSubset = (subset: Range, superset: Range): boolean => {
...
+  ranges: Range[],
+  rangeToSpan: Range
+): boolean => {
+  const sortedRanges: Range[] = ranges.sort((a: Range, b: Range) => {
+    return a.start.lt(b.start) ? -1 : a.start.eq(b.start) ? 0 : 1
+  })
+
+  let spannedEnd: BigNum = rangeToSpan.start
```
akhila-raju: 'Spannedend' as well I'm struggling to understand the meaning here

packages/core/src/app/ovm/state-manager.ts
willmeister: Yep 😄

akhila-raju: oooh I didn't know this 😄!

packages/core/src/app/utils/range.ts
willmeister: Good call -- I updated the name of this var and added some comments and examples to the function to make it a bit clearer what it's trying to do.

packages/core/src/app/utils/range.ts
```diff
@@ -66,7 +66,11 @@ export const isRangeSubset = (subset: Range, superset: Range): boolean => {
 }
 
 /**
- * Determines whether the provided Ranges collectively span the Range in question
+ * Determines whether the provided Ranges collectively span the Range in question.
+ * For instance,
+ * doRangesSpanRange([{start: 1, end: 3}, {start: 3,  end: 5}], {start: 2, end: 4})
```
akhila-raju: Actually, adding the 'do' makes more sense! The function name needs to be updated as well to correspond with the function name in the comment :)

akhila-raju: I like the comments that were added, definitely a lot easier to understand, thanks!

packages/core/src/app/utils/crypto.ts
```diff
@@ -13,5 +13,32 @@ export const verifySignature = (
...
+}
+
+/**
+ * Decrypts the provided encrypted message with the provided public key
+ *
+ * @param publickey the public key in question
+ * @param encryptedMessage the encrypted message to decrypt
+ */
+export const decryptWithPublicKey = (
+  publickey: any,
```
karlfloersch: This should be a `privatekey` I believe

packages/core/src/app/utils/equals.ts
```diff
@@ -1,5 +1,11 @@
 import { Range } from '../../types/db'
-import { StateObject, StateUpdate } from '../../types/serialization'
+import {
+  BlockTransaction,
+  StateObject,
+  StateUpdate,
+  Transaction,
+} from '../../types/serialization'
+import { unwatchFile } from 'fs'
```
karlfloersch: Is the `unwatchFile` import used?

packages/core/src/types/aggregator/aggregator.interface.ts
```diff
@@ -0,0 +1,23 @@
...
+   *
+   * @param transaction the Transaction in question
+   * @param witness the Witness of the Transaction in question
+   *
+   * @returns the BlockTransactionCommitment indicating the transaction will be included in the next block
+   */
+  ingestTransaction(
+    transaction: Transaction,
+    witness: string
+  ): Promise<BlockTransactionCommitment>
```
karlfloersch: It is unclear to me what the `witness` parameter does here. I believe it may have been part of our out of date spec.

Ideally we'd simply be sent the transaction itself. From this we run some authentication logic & generate a new StateUpdate. In fact this whole transaction ingestion process is under specified and so I just created an issue for it here: https://github.com/plasma-group/pigi/issues/349

packages/core/src/app/ovm/state-manager.ts
```diff
@@ -24,13 +24,10 @@ import { getOverlappingRange, ONE, rangesIntersect } from '../../app'
...
-  private stateDB: StateDB
-  private pluginManager: PluginManager
-
-  public constructor(stateDB: StateDB, pluginManager: PluginManager) {
-    this.stateDB = stateDB
-    this.pluginManager = pluginManager
-  }
+  public constructor(
+    private stateDB: StateDB,
+    private pluginManager: PluginManager
```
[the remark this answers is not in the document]
willmeister: Yeah, I think you're right. Does that make sense to do in a separate PR that OVM-ifies all of our outdated names?

packages/core/src/app/utils/crypto.ts
willmeister: I think I was imagining that this would simply get the message that was signed by the private key associated with the provided `publicKey`... basically doing what `verifySignature` does, but returning the original message. 

That said, I imagine we'll also need the reverse: decrypting a message with a private key that was previously encrypted with the corresponding public key.

packages/core/src/app/utils/equals.ts
willmeister: Haha, nooooope 😄

packages/core/src/types/aggregator/aggregator.interface.ts
willmeister: That's a good question. There is both authentication and authorization to be done, right?

1. Is this transaction actually from the user who is trying to send their funds?
2. Does the user trying to send funds actually own the funds they're trying to send?

I assumed the witness portion would solve (1), although that could be handled in the communication layer itself. Is there something I'm missing that either handles that or makes it unnecessary?

packages/core/src/app/utils/crypto.ts
karlfloersch: Ahh got it

packages/core/src/types/aggregator/aggregator.interface.ts
karlfloersch: I think that we can keep transactions more general and with that we won't need to break those two (authentication/authorization) into two steps. Instead you simple check that the transaction itself is valid. The tx might not even have a single sender--it could simply be a subsidized function call in a smart contract (essentially the concept of account abstraction https://github.com/ethereum/EIPs/issues/859).

> although that could be handled in the communication layer itself
Yeah I think it makes sense for the anti-spam mechanism to be handled by the communication layer. Eg. banning peers that repeatedly broadcast failing transactions.

packages/core/src/app/ovm/state-manager.ts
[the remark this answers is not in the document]
karlfloersch: Yeah let's do in a separate PR

packages/core/src/types/aggregator/aggregator.interface.ts
willmeister: Thanks for the explanation 😄. I'll remove witness from the method signature.

PR #357, opened 2019-07-23
One BigNumber to Rule Them All

### LINKED ISSUE #287: Resolve BigNumber/BN usages
- Export BigNum as a file so that naming doesn't change
- Figure out a clean way to wrap the ether.js bigNum library so that we always are working in bn.js

Right now, because we're using a mix of web3 libraries which use two different BigNum libraries, dealing with encoding/decoding has involved a lot of stringification and re-casting.  It's confined for now, but could be a real pain going forward.

Ideally, we would look at all the places where these libraries are being used, and choose one consistent one internally, only doing re-casting in the calls we need.

**Describe alternatives you've considered**
In the worst case, we could write an agnostic wrapper class which serves each when needed.

**Additional context**
It's probably best to do this at the same time as we build a more robust encoding/decoding library.

willmeister:
# Description
Adds BigNumber class to wrap big number implementation and allow changes without breaking dependencies.

### Fixes
- Fixes #287 

## Contributing Agreement

### DIFF
--- a/packages/core/src/app/db/range-bucket.ts
+++ b/packages/core/src/app/db/range-bucket.ts
@@ -3,9 +3,6 @@
  * Credit to the original author, Christopher Jeffrey (https://github.com/chjj).
  */
 
-/* External Imports */
-import BigNum = require('bn.js')
-
 /* Internal Imports */
 import {
   Bucket,
@@ -17,10 +14,17 @@ import {
   V,
   RangeBucket,
   RangeEntry,
-  Endianness,
   PUT_BATCH_TYPE,
 } from '../../types'
-import { bufferUtils, intersects, BaseDB } from '../../app'
+import {
+  bufferUtils,
+  intersects,
+  BaseDB,
+  BigNumber,
+  BIG_ENDIAN,
+  Endianness,
+  ZERO,
+} from '../../app'
 
 /* Logging */
 import debug from 'debug'
@@ -42,7 +46,7 @@ export class BaseRangeBucket implements RangeBucket {
     readonly db: DB,
     readonly prefix: Buffer,
     readonly keyLength: number = 16,
-    readonly endianness: Endianness = 'be'
+    readonly endianness: Endianness = BIG_ENDIAN
   ) {}
 
   /**
@@ -62,7 +66,7 @@ export class BaseRangeBucket implements RangeBucket {
    * @param value A Buffer value, likely to be stored.
    * @returns resulting concatenation of the start key & input value
    */
-  private addStartToValue(start: BigNum, value: Buffer): Buffer {
+  private addStartToValue(start: BigNumber, value: Buffer): Buffer {
     return Buffer.concat([
       start.toBuffer(this.endianness, this.keyLength),
       value,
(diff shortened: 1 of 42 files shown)


packages/core/test/app/utils/number.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+      assert(num.toNumber() === 10, 'Could not construct BigNumber with number')
+    })
+
+    it('should handle BigNumber', () => {
+      const num: BigNumber = new BigNumber(new BigNumber(10))
+      assert(num.toNumber() === 10, 'Could not construct BigNumber with number')
+    })
+
+    it('should handle Buffer', () => {
+      const num: BigNumber = new BigNumber(new BigNumber(10).toBuffer())
```
karlfloersch: For completeness we could explicitly test `toBuffer()`

packages/core/test/app/utils/number.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+    //   assert(new BigNumber(100).pow(new BigNumber(.5)).eq(new BigNumber(10)))
+    // })
+    //
+    // it('test negative power', () => {
+    //   assert(new BigNumber(100).pow(new BigNumber(-2)).eq(new BigNumber(0.0001)))
+    // })
+    //
+    // it('test negative fractional power', () => {
+    //   assert(new BigNumber(100).pow(new BigNumber(-.5)).eq(new BigNumber(0.1)))
+    // })
```
karlfloersch: Also bn.js does not support `log()` for BigNumbers :'(

packages/core/test/app/utils/number.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+      assert(new BigNumber(11).compare(new BigNumber(10)) === 1)
+      assert(new BigNumber(-10).compare(new BigNumber(-11)) === 1)
+    })
+
+    it('test equal', () => {
+      assert(new BigNumber(10).compare(new BigNumber(10)) === 0)
+      assert(new BigNumber(-10).compare(new BigNumber(-10)) === 0)
+    })
+  })
+})
```
karlfloersch: Considering how comprehensive these tests are it is worth noting that some `toX()` functions were not tested. In particular toBuffer, toJSON, & toString().

packages/core/test/app/utils/number.spec.ts
willmeister: Yeah, that may be a good idea. I'm less concerned with what is present in the buffer and more concerned that when we call `.toBuffer()` and then instantiate the result, we get the same value. Might be good to test the different bases, though.

packages/core/src/app/db/range-bucket.ts
```diff
@@ -122,13 +122,13 @@ export class BaseRangeBucket implements RangeBucket {
...
    */
-  private validateRange(start: BigNum, end: BigNum): void {
+  private validateRange(start: BigNumber, end: BigNumber): void {
     // Make sure start is less than end
     if (!start.lt(end)) {
       throw new Error('Start not less than end')
     }
     // Make sure start is greater than or equal to zero
-    if (!start.gten(0)) {
+    if (!start.gte(ZERO)) {
```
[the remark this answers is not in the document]
willmeister: The main reason that `ZERO` and `ONE` exist, other than convenience, is to avoid allocating memory over and over again for very common numbers. My thinking is that there shouldn't be any naming conflicts with other types of numbers because all other numbers that we would use should be primitives, so there's no benefit in creating constants for them. Happy to rename if you think I should, though 😄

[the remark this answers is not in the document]
karlfloersch: > all other numbers that we would use should be primitives

Ahh good point :)

PR #365, opened 2019-07-24
[WIP] Figuring Out Decider Persistence and HashPreimageExistenceDecider

### LINKED ISSUE #336: Implement HashPreimageExistenceDecider
- HashPreimageExistenceInput (includes a hash image)
    - HashPreimageExistenceWitness
    - Implement decide() & checkDecision()

### LINKED ISSUE #362: Design/Implement Decision Storage
Figure out the best way to inject a LevelDB into deciders & organize this in logical buckets that don't overlap (eg. access the wrong bucket).

willmeister:
## Description
Proposes a base class for all Deciders that store data in a DB, and provides an example implementation in HashPreimageExistenceDecider

## Questions
- Will nested serialization / deserialization be a huge pain? If so, is there anything we can do about it?
- Does this interface seem reasonable for DbDecider? Is it generic enough? Does it take any power away from the extension classes or stay lightweight and just simplify things?

### Fixes
- Fixes #336 
- Fixes #362 

## Contributing Agreement


packages/core/src/app/ovm/deciders/db-decider.ts
```diff
@@ -0,0 +1,70 @@
...
+  public async checkDecision(input: any): Promise<Decision> {
+    const hash: Buffer = DbDecider.getCacheKey(input)
+    const decisionBuffer: Buffer = await this.decisionBucket.get(hash)
+
+    if (decisionBuffer === null) {
+      return undefined
+    }
+
+    return this.deserializeDecision(decisionBuffer)
+  }
```
karlfloersch: Hmm does this only work for decisions on particular keys? For for instance, I don't think this works if we want a decision over a range.

packages/core/src/app/ovm/deciders/db-decider.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Stores the provided decision for the provided input.
+   *
+   * @param input the input that resulted in the provided decision
+   * @param serializedDecision the buffer representing the Decision to be stored
+   */
+  protected async storeDecision(
+    input: any,
+    serializedDecision: Buffer
+  ): Promise<void> {
+    const key: Buffer = DbDecider.getCacheKey(input)
```
karlfloersch: Once again don't think this key works for ranges

packages/core/src/app/ovm/deciders/db-decider.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  public abstract decide(_input: any, _witness: any): Promise<Decision>
+
+  /**
+   * Returns the unique ID of this Decider.
+   *
+   * This is used to identify this Decider in storage and serialization / deserialization
+   * @returns the unique ID
+   */
+  protected abstract getUniqueId(): string
```
karlfloersch: Nice

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
```diff
@@ -0,0 +1,111 @@
...
+   * @returns the Decision
+   */
+  private constructDecision(
+    hash: string,
+    outcome: boolean,
+    preimage: string
+  ): Decision {
+    return {
+      outcome,
+      implicationProof: [
```
karlfloersch: I like calling this `justification` -- this way it's pretty clear. You call `checkDecision(...)` and then it returns a decision + a justification for why that decision was made.

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
```diff
@@ -0,0 +1,111 @@
...
+        {
+          implication: {
```
karlfloersch: Hmm... would it make sense to have the data structure be: 
```
interface Decision {
  property: Property,
  justification: witness | Decision[],
}
```

packages/core/src/app/ovm/deciders/db-decider.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * Gets the unique key for the provided input to use as a cache key for its Decisions
+   *
+   * @param input the input for which a key will be computed
+   * @returns the computed cache key
+   */
+  private static getCacheKey(input: any): Buffer {
+    return Buffer.from(Md5.hashStr(JSON.stringify(input)) as string)
```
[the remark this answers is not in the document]
willmeister: Yeah, can certainly do that. Could also change it here down the line. Really just looking for something reasonably fast that produces unique finite-sized output. I'll pull it out into crypto.ts and/or have it take in a hash function as a dependency.

packages/core/src/app/ovm/deciders/db-decider.ts
willmeister: Yeah, this will likely need to be overridden at times. If the `Decision` is directly associated with the `input` this should work, but in the case where the `Decision` may be associated with _another input_ that is derived from the input, it will not work.

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
willmeister: Changed the name to `justification`, but do we want to change the type? Can't remember where we landed.

packages/core/src/app/ovm/deciders/db-decider.ts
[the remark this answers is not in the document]
willmeister: For transparency, I kept it as non-configurable because there's no reason an implementation should need to configure it (or even know how keys are generated).

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
willmeister: Also do _all_ of the contained `Decision`s have an `outcome`?.

packages/core/src/app/ovm/deciders/db-decider.ts
```diff
@@ -0,0 +1,69 @@
+import { Decider, Decision } from '../../../types/ovm/decider.interface'
+import { Bucket, DB } from '../../../types/db'
+import { Md5Hash } from '../../utils'
+
+export abstract class DbDecider implements Decider {
```
karlfloersch: We could consider calling this a `KVDecider` ? Because specifically it specializes in key value pairs

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
karlfloersch: > but do we want to change the type?

I can't remember either, so maybe for now we stick with it & see where it breaks?

> Also do all of the contained Decisions have an outcome?.

I think so.

packages/core/src/app/ovm/deciders/hash-preimage-existence-decider.ts
```diff
@@ -0,0 +1,119 @@
...
+    witness: PreimageWitness,
+    input: HashInput,
+    outcome: boolean
+  ): Buffer {
+    return Buffer.from(
+      JSON.stringify([
+        witness.preimage.toString(),
+        input.hash.toString(),
+        outcome,
+      ])
```
karlfloersch: This seems like a pretty good serialization technique! I wish we had more "standard" ways to serialize, but I don't know if it's really that possible & I like this function's simplicity.

packages/core/test/app/ovm/deciders/hash-preimage-existence-decider.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+      const justification: ImplicationProofElement = decision.justification[0]
+      justification.implication.decider.should.equal(decider)
+      justification.implication.input['hash'].should.equal(hash)
+      justification.implicationWitness['preimage'].should.equal(preimage)
+    })
+
+    it('should throw invalid preimage', async () => {
+      try {
+        await decider.decide({ hash }, { preimage: Buffer.from('womp womp') })
```
karlfloersch: 'womp womp' 😂

packages/core/test/app/ovm/deciders/hash-preimage-existence-decider.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      } catch (e) {
+        if (e instanceof CannotDecideError) {
+          assert(true, 'This is expected to happen')
+        } else {
+          assert(false, 'CannotDecideError was expected.')
+        }
```
karlfloersch: I think you may be able to use https://www.chaijs.com/api/bdd/#method_throw instead of the `try {} catch {}`. That said, I personally don't remember using `throw()` with an async function so it could be a little more tricky.

That said, I do like the 'low tech-ness' of your current approach so take this as just an alternative not so much a suggestion.

packages/core/test/app/ovm/deciders/hash-preimage-existence-decider.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      }
+    })
+  })
```
karlfloersch: Another test could be making two preimage decisions

packages/core/test/app/ovm/deciders/hash-preimage-existence-decider.spec.ts
willmeister: Gotta have fun with tests 😉

PR #375, opened 2019-07-29
And & Not decider implementation

### LINKED ISSUE #332: Implement AND Decider
- Accept two properties & decide `true` if both of those properties are decided true. Decides `false` if either `false`, decides `UNDECIDED` if neither false and at least one `UNDECIDED`
    - ANDInput
    - ANDImplicationWitness (= `left` or `right`, whichever returned false)

### LINKED ISSUE #333: Implement NOT Decider
- NOTDeciderInput
    - Accepts a property & decides `true` if that property is decided false. Decides `UNDECIDED` if the property returns `UNDECIDED`.

willmeister:
Implements simple AND and NOT deciders that return the AND of two different passed in properties and the NOT of a single passed in property, respectively.

## Questions
- It makes sense for these to not persist anything, right? I'm thinking they should just rely on the storage of the underlying deciders.
- Does the input look right? The other option is to have the witness(es) passed as the witness param, but then checkDecision doesn't work without persistence.

### Fixes
- Fixes #332
- Fixes #333

## Contributing Agreement


packages/core/src/app/ovm/deciders/and-decider.ts
```diff
@@ -0,0 +1,52 @@
...
+      input.left.decider.decide(input.left.input, input.leftWitness),
+      input.right.decider.decide(input.right.input, input.rightWitness),
+    ])
+
+    if (!leftDecision.outcome) {
+      return leftDecision
+    }
+    if (!rightDecision.outcome) {
+      return rightDecision
+    }
```
karlfloersch: Here I believe instead of only returning the decision which invalidated it, we need to return a new decision with this AND included as an implication proof element. This is so that when we guard the justification for our local decision includes the AND decision

packages/core/src/app/ovm/deciders/and-decider.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const justification: ImplicationProofElement[] = []
+    if (!!leftDecision.justification.length) {
+      justification.concat(leftDecision.justification)
+    }
+    if (!!rightDecision.justification.length) {
+      justification.concat(rightDecision.justification)
+    }
```
karlfloersch: So we’ve commonly said that the `justification` will only be used when we decide `false`. This has been said because the case we were considering was when the on-chain decision differs from our local decision and we want to guard.

That said after looking at this it’s clear that when we want to send a coin & need to transfer a “proof” that this justification will come in very handy as it should be the entire proof that we are going to end up sending. Cool!

packages/core/src/app/ovm/deciders/not-decider.ts
```diff
@@ -0,0 +1,37 @@
...
+    witness: undefined
+  ): Promise<Decision> {
+    const decision: Decision = await input.property.decider.decide(
+      input.property.input,
+      input.witness
+    )
+
+    if (!decision) {
+      decision.outcome = true
+      return decision
```
karlfloersch: I think here the return value should be:
```
{
  outcome: true,
  justification: [decision]
}
```

packages/core/src/app/ovm/deciders/not-decider.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    decision.outcome = false
+    return decision
```
karlfloersch: Similarly here

packages/core/src/app/ovm/deciders/and-decider.ts
willmeister: Yeah, I tend to agree. To say that it's only necessary when we decide false is to say that we know how it's going to actually be used up the chain, and I don't think we do.

packages/core/test/app/ovm/deciders/and-decider.spec.ts
```diff
@@ -0,0 +1,283 @@
+import '../../../setup'
+
+import {
+  AndDecider,
+  AndDeciderInput,
+  CannotDecideError,
+} from '../../../../src/app/ovm/deciders'
+import { CannotDecideDecider, FalseDecider, TrueDecider } from './utils'
```
karlfloersch: These are great

packages/core/test/app/ovm/deciders/and-decider.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      await testThrowWhenLeftCannotDecide(false)
+    })
+
+    it('should throw when right cannot decide', async () => {
+      await testThrowWhenRightCannotDecide(false)
+    })
+
+    it('should throw when both cannot decide', async () => {
+      await testThrowWhenBothCannotDecide(false)
+    })
```
karlfloersch: These test cases seem pretty comprehensive! Somewhat strange here how similar `decide(...)` and `checkDecision(...)` are, but still think it all makes sense at this point.

The only other tests I could think of were: 1) bad input values, and 2) nested AND predicates. That said, don’t think either are critical & personally would be happy with the current level of test coverage.

packages/core/test/app/ovm/deciders/not-decider.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      await testReturnTrueWithFalseDecision(false)
+    })
+
+    it('should return false with true decision', async () => {
+      await testReturnFalseWithTrueDecision(false)
+    })
+
+    it('should throw when cannot decide', async () => {
+      await testThrowWhenCannotDecide(false)
+    })
```
karlfloersch: Coverage once again LGTM :)

PR #379, opened 2019-08-01
SignedByQuantifier

### LINKED ISSUE #340: SignedByQuantifier
- analagous to the inteval tree intersection quantifier, in that it interacts with the db used by `SignedByDecider`
    - communicates with wallet so that a `SignedByQuantifier` can check whether the `input.signer == MY_ADDRESS`, allowing it to return `true` for `isCertainlyAll`

willmeister:
## Description
Implemented the SignedByQuantifier which quantifies all messages signed by a specific address.

## Questions
- JSON DB was a bit rushed, do you think this is the best way to store Messages and other unstructured data

### Fixes
- Fixes #340 

## Contributing Agreement


willmeister:
Included in #382

PR #378, opened 2019-07-31
SignedByDecider

willmeister:
## Description
SignedByDecider that determines whether or not a signature is the result of a message signed by the private key associated with a provided public key.

### Fixes
- Fixes #339 

## Contributing Agreement

- [x] I have read and understood the [Plasma Group Contributing Guide and Code of Conduct](https://github.com/plasma-group/pigi/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #382, opened 2019-08-07
State Channels (Payment Channels)

### LINKED ISSUE #337: Test Preimage Existence on a Range of Hash Images
This is an integration test that tests a bunch of other features end-to-end:
    - uses IntegerRangeQuantifier and the PreimageExistenceDecider to make decisions on ranges of preimages
    - uses a PropertyFactory which contains a list of `hashImages: bytes32[]` and constructs a PreimageExists Property with input as `hashImages[result]` via `for (result of IntegerRangeQuantifier.getAllSuchThat(...).results)`

### LINKED ISSUE #339: Implement SignedByDecider
SignedByDecider:
    - allows us to verify existence of and store a signature by some pubkey.
    - `type SignedByInput = {pubkey: address, message: bytes}`
    - type `type SignedByWitness = {v, r, s}` (standard ECDSA signature type)
    - stores in a DB indexed by `[pubkey][message]`

Note: If possible, should have a signature checker injected into it and have a general witness representation so it can use a variety of signing mechanisms.

### LINKED ISSUE #340: SignedByQuantifier

### LINKED ISSUE #341: Test Simple State Channel
Use SignedByDecider and/or SignedByQuantifier to create StateChannel functionality as an example OVM implementation.

willmeister:
## Description
Creates the deciders and logic necessary for State Channels (really payment channels, but close enough, right?). Adds unit tests demonstrating State Channel message functionality.

## Questions
- Do we have to serialize and deserialize the full State Channel claim, or is having a contract hash in the message a good enough proxy?

### Fixes
- Fixes #341 
From other PRs merged into this one:
- Fixes #340 
- Fixes #339 
- Fixes #337 

## Contributing Agreement


packages/core/src/app/ovm/deciders/signed-by-decider.ts
```diff
@@ -0,0 +1,137 @@
...
+        input,
+        SignedByDecider.serializeDecision(witness, input, signatureMatches)
+      )
+    }
+
+    return this.constructDecision(
+      witness.signature,
+      input.publicKey,
+      input.message,
+      signatureMatches
```
ben-chain: It is not quite the case that, if the signature doesn't match, that we can decide "false"--we can only fail to make the decision.  Intuitively, "just because this signature isn't right doesn't mean the right one isn't out there."

packages/core/src/app/ovm/deciders/there-exists-such-that-decider.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  private getDecision(
+    input: ThereExistsSuchThatInput,
+    trueDecision: Decision,
+    falseDecisions: Decision[],
+    undecided: boolean
+  ): Decision {
+    if (!trueDecision && undecided) {
+      throw new CannotDecideError(
+        'Cannot decide ForAllSuchThat due to undecided Decision or not all results being quantified.'
```
ben-chain: typo: `ForAllSuchThat`-->`ThereExistsSuchThat`

packages/core/src/app/ovm/examples/simple-state-channel.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    return undefined
+  }
+
+  private async disputeMessage(
+    message: ParsedMessage,
+    claim: Property,
+    witness: any
+  ): Promise<void> {
+    // TODO: build dispute that will dispute StateChannelClaim
```
ben-chain: This is definitely not a priority for the purposes of this state channel PoC, but note that it should not be needed in the long term as the OVM will auto-handle disputes whenever it sees a claim it can decide false, based on the justification.  Before then, I think the right place to unit test disputing a state channel claim will be the contract tests, so no need to worry about it for now :)

packages/core/src/app/serialization/examples/simple-state-channel.ts
```diff
@@ -0,0 +1,81 @@
...
+  decider: OrDecider
+  input: {
+    properties: [
+      {
+        decider: ThereExistsSuchThatDecider
+        input: {
+          quantifier: SignedByQuantifier
+          quantifierParameters: any
+          propertyFactory: MessageInvalidatedClaim
+          witnessFactory: MessageInvalidatedWitnessFactory
```
ben-chain: A little confused by there being a witness here in the input.  Generally witnesses are used to make a decision for some property, but the witness is not a part of the input to that property.

packages/core/src/app/ovm/deciders/or-decider.ts
```diff
@@ -0,0 +1,122 @@
...
+      throw new CannotDecideError(
+        'At least one of the OR deciders could not decide and none returned true, so this cannot be decided.'
+      )
+    }
+
+    return this.getDecision(input, {
+      outcome: false,
+      justification: falseJustifications,
+    })
+  }
```
karlfloersch: This decider looks good to me! The only comment would be that most of the time I imagine we will often end up using NOT NOT AND in place of OR, but that doesn’t change the fact that this looks solid

packages/core/src/app/ovm/deciders/examples/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+    return (
+      message.message.channelId.equals(other.message.channelId) &&
+      message.message.nonce.equals(other.message.nonce) &&
+      (message.sender.equals(other.sender) ||
+        message.sender.equals(other.recipient)) &&
+      (message.recipient.equals(other.recipient) ||
+        message.recipient.equals(other.sender)) &&
+      !objectsEqual(message.message.data, other.message.data)
+    )
+  }
```
karlfloersch: Maybe this function should be `stateChannelMessagesConflict(...)`

packages/core/src/app/ovm/deciders/signed-by-decider.ts
```diff
@@ -0,0 +1,137 @@
...
+    input: SignedByInput,
+    outcome: boolean
+  ): Buffer {
+    return Buffer.from(
+      JSON.stringify([
+        witness.signature.toString(),
+        input.publicKey.toString(),
+        input.message.toString(),
+        outcome,
+      ])
```
karlfloersch: Could technically use your nice serialization lib `objectToBuffer(...)`

packages/core/src/app/ovm/examples/simple-state-channel.ts
```diff
@@ -0,0 +1,214 @@
...
+    const signatures = {}
+    signatures[signedMessage.sender.toString()] = signaturePlaceholder
+    return {
+      sender: signedMessage.sender,
+      recipient: this.myAddress,
+      message,
+      signatures,
+    }
+  }
+}
```
karlfloersch: This is great

packages/core/src/app/serialization/examples/simple-state-channel.ts
```diff
@@ -0,0 +1,81 @@
...
+      decider: ForAllSuchThatDecider
+      input: {
+        quantifier: SignedByQuantifier
+        quantifierParameters: any
+        propertyFactory: MessageInvalidatedOrCurrentPropertyFactory
+        witnessFactory: WitnessFactory
+      }
+    }
+  }
+}
```
karlfloersch: Ben may have already left comments here but this claim construction can likely be simplified using the “does not exist” approach — turns out you can add a concept of time to a does not exist & thereby say “it does not exist @ block/timestamp/whatever” which of course we like for caching.

packages/core/src/app/ovm/deciders/signed-by-decider.ts
willmeister: Ahh ok perhaps I misunderstood the motivation for that decider -- I had assumed that it was to determine if X was signed by Y, not whether or not there exists some Y that has signed X. Looking at the input vs witness, I see that your point is totally valid.  This is pulled in from the other PR ( #378 ) since it's a dependency and that one isn't approved, so I'll make sure to add it there for reference.

packages/core/src/app/ovm/deciders/and-decider.ts
```diff
@@ -16,6 +16,15 @@ export interface AndDeciderInput {
  * Decider that decides true iff both of the provided properties evaluate to true.
  */
 export class AndDecider implements Decider {
+  private static _instance: AndDecider
```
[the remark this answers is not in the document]
willmeister: Ohh, yeah, it's not. Was thinking that it might make sense for these to be singletons, and it was useful in some code that was deleted, but this can be removed.

packages/core/src/app/ovm/deciders/or-decider.ts
willmeister: Yeah, I just made the OR to reduce nesting and [originally] I needed to support 3 sub-deciders, and AND only supports 2, so I made OR and made it support any number of inputs.

packages/core/src/app/ovm/deciders/there-exists-such-that-decider.ts
willmeister: Copy / paste 😞

packages/core/src/app/serialization/examples/simple-state-channel.ts
willmeister: That definitely makes sense. The witness factory is just a mechanism to pass witnesses to nested deciders. It's not a witness per se, but a function that will generate a witness in the context in which it's used, but your point is still valid.

Either way, I'm interested to hear your thoughts if you think we can make this better!

packages/core/src/app/ovm/deciders/signed-by-decider.ts
willmeister: Turns out you can't override Buffers' serialization since the Buffer class has a `toJSON(...)` method. That might mark the end of `objectToBuffer` =/

packages/core/test/app/ovm/examples/simple-state-channel.spec.ts
```diff
@@ -0,0 +1,729 @@
...
+        parsedMessage.message.channelId,
+        [bAddress],
+        parsedMessage.message.data['addressBalance']
+      )
+
+      await acknowledgeMessage(nextMessage, a, new BigNumber(2))
+    })
+  })
+
+  describe('channel exit', () => {
```
ben-chain: This section is the super duper awesome meat of the OVM work here--epic!!!

I think generally, it might be useful to in the future to explicitly/manually construct the exit claims which are being checked behind the scenes here.  Since these decisions are being outsourced to the state channel client, it's a little tricky to parse what's going on between these two files--so it might've been helpful to first write a manual exit decision in the test, so that a reader can follow what's going on "closer to the metal."

But, since this is an example, it's not critical, and not worth redoing that work now--just something to keep in mind for the future. 😃

packages/core/src/app/ovm/deciders/examples/utils.ts
willmeister: It's probably not going to be used for other messages, so you're right.

In principle, this should work for any type that follows the base `Message` type. The contents of `data` is what determines the type of message (like `StateChannelMessage`), and we don't make any assumptions on that in this function.

packages/core/test/app/ovm/examples/simple-state-channel.spec.ts
willmeister: Good call. This definitely makes sense to isolate the test as well. I'll update.

packages/core/src/app/ovm/deciders/examples/utils.ts
willmeister: Updated for good measure 😄

PR #384, opened 2019-08-14
DataStore-backed Deciders in place of witnesses

### LINKED ISSUE #381: Stand-alone Data Stores used by Deciders
This ticket encompasses:
* Creating individual data stores for necessary `Decider` data
* Injecting these stores into `Deciders`
* Updating references in unit tests
* Thinking through (and implementing if possible) a way to let all data stores know about all incoming and outgoing messages so they may store any pertinent info
* *Most Importantly* This should eliminate the need for a witness (only for these specific deciders)

Details [here](https://hackmd.io/ycrtl9CNT12S3TVC_7z8fQ).

willmeister:
## Description
Creating stand-alone datastores to be used by some deciders so that they can be used elsewhere to store and access data on which deciders decide.

## Questions
- Thoughts on approach, interfaces, etc?

### Fixes
- Fixes #381 

## Contributing Agreement


ben-chain:
Sorry for taking so long to review!  1 main question:
Looking at the code, it seems like the `SignedByQuantifier` is not using the same `SignedByDb` as the corresponding `SignedByDecider`.  Is that correct?  If so, we should change the quantifier to make them the same--I think a big reason to do it this way is that they can share a DB.

Another question--are there any cases with this change where `witness` is used in the decide function?  I think a goal is for it not to be, and be able to remove it.  For example when we "decide" a preimage exists, this happens implicitly by sending it to the `preImageDB`, and from there on future queries to `decide` will return true as a result.  That way, we never have to pass the preimage directly into the decider.

PR #403, opened 2019-08-23
Sparse Merkle Tree & MerkleInclusionQuantifier

### LINKED ISSUE #386: Implement MerkleInclusion Quantifier
We need a quantifier which functions as the $dataIncludedAt$ in the above claim. Particularly, it would return some `includedData` included at a particular point in a vanilla Merkle tree (above, each are called $inputs_i$) for `getAllQuantified`, or none if an exclusion proof is relevant.

Specifically, we would have 
```typescript=
type MerkleIntervalInclusionParams =
    {
        root: bytes32,
        key: uint32
    }
```
The `getAllQuantified` would return some generic `data: bytes` to be fed to a property factory.

Note that the length of `key` above should be parametrizable, i.e. 20 bytes for an address or 32 if we want to be verbose.  I think it's too complex and maybe infeasible to have different-length keys in a single tree, though.

willmeister:
## Description
Implements a Sparse Merkle Tree and MerkleInclusionQuantifier using it.

### Fixes
- Fixes #386 

## Contributing Agreement


karlfloersch:
> Do we need / want a DB in SMT? If nothing else, we'll need to store nodes in memory, so we can just create an in memory DB, so might as well use a DB, right?

Yep! DB is the way to go. We can for sure use an in-memory DB if we need extra performance so you’re totally right that we might as well implement as a DB. The state will likely grow very large so some sort of in-memory cache could make sense even if in the long term when we have too much state to fit in RAM.

karlfloersch:
Oh btw one thing that would be very interesting is if you could benchmark this! Basically insert random keys using a tree of height 160 or so & see how many inserts we can do per second

packages/core/src/types/utils/merkle-tree.interface.ts
```diff
@@ -0,0 +1,56 @@
+import { BigNumber } from '../number'
+
+export interface MerkleTreeNode {
+  key: BigNumber
```
ben-chain: Just to confirm--this is both the depth and position of the node in the tree?

packages/core/src/app/ovm/quantifiers/merkle-inclusion-quantifier.ts
```diff
@@ -0,0 +1,36 @@
...
+  root: Buffer
+  key: BigNumber
+}
+
+/*
+ * The MerkleInclusionQuantifier that quantifies data at a specific key of a Merkle Tree,
+ * if data exists at the key.
+ */
+export class MerkleInclusionQuantifier implements Quantifier {
+  public constructor(private readonly tree: MerkleTree) {}
```
ben-chain: So it's interesting that we pass a tree into the constructor, but then also we query the `root`--as intuitively "a merkle tree' has just one unique root.  I guess this is because the `MerkleTree` values can be updated.  I wonder if we should specify that the tree class is dynamic?

packages/core/src/types/utils/merkle-tree.interface.ts
```diff
@@ -0,0 +1,56 @@
...
+   * including all ancestors' hashes that result from this modification.
+   *
+   * @param key The key to update
+   * @param value The new value
+   * @return true if the update succeeded, false if we're missing the intermediate nodes / siblings required for this
+   */
+  update(key: BigNumber, value: Buffer): Promise<boolean>
+
+  /* TODO when we need it
+   * getMerkleProof(key: BigNumber, value: Buffer): Promise<MerkleTreeInclusionProof>
```
ben-chain: Just as a note, this will be important to have when integration testing with the contracts, so will be a priority before then.  Let's make a ticket after merging this?

packages/core/test/app/ovm/quantifiers/merkle-inclusion-quantifier.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        result.results[0].equals(value),
+        `Result value [${result.results[0].toString()}] found when expecting [${value.toString()}]`
+      )
+      assert(
+        result.allResultsQuantified,
+        'All results should be quantified if we are able to find the leaf node in the tree.'
+      )
+    })
+
+    it('returns value on populated tree tree with a root that matches but mismatched key', async () => {
```
ben-chain: "returns value"--should this be "returns undefined" or am I missing something?

packages/core/src/types/utils/merkle-tree.interface.ts
willmeister: No. It does not contain depth information. Bits after the bit that indicates whether this node is the left or right child of its parent will exist but will all be zeros. This makes it equivalent to all keys for all left descendants of this node as well.

packages/core/src/app/ovm/quantifiers/merkle-inclusion-quantifier.ts
willmeister: The `MerkleTree` interface specifies that itself, no?

packages/core/test/app/ovm/quantifiers/merkle-inclusion-quantifier.spec.ts
willmeister: Yes, thanks for reading thoroughly 😄

packages/core/src/types/utils/merkle-tree.interface.ts
willmeister: Alternatively, we could right-shift it so that the least significant bit indicates whether this node is the left or right child of its parent, but 1) we'd have to left shift it back to actually walk the tree and 2) we'd have the same problem in that the key for 0 is the same as 00, which is the same as 000 in that they'd all just be the `BigNumber(0)`

ben-chain: Got it--I trust your approach, was just trying to better understand what the "key" of a non-leaf node is. 😃

packages/core/src/app/ovm/quantifiers/merkle-inclusion-quantifier.ts
ben-chain: Oh totally--I guess I was just saying we could consider naming the interface `UpdateableMerkleTree` or something of the like.  Not sure it's worth the verbosity though, you're right that that is apparent when you check the interface (which is pretty concise)

packages/core/src/types/utils/merkle-tree.interface.ts
willmeister: Ohh, gotcha. It's just used to de-dupe in storage. Before ID in the DB was just the hash, but that wasn't sufficient because if we stored two intermediate nodes at different spots in the tree that happen to have the same hash, we don't want an update to one to delete the other and vice-versa. This solves that: https://github.com/willmeister/pigi/blob/feat/386/MerkleInclusionQuantifier/packages/core/src/app/utils/merkle-tree.ts#L556

ben-chain: Right, I remembered this convo and that's why I assumed it was depth and position--but that solution makes sense.

PR #405, opened 2019-08-25
Add Rollup Merkle Tree Utils

### LINKED ISSUE #395: Implement `RollupChain` contract
Implement the RollupChain contract which, for now, has a `submitBlock(...)` function as well as `proveInvalidTransition(...)` function. Later on these will be split up, but we can shove it all together for now.

Needs to emit an event every time a block is proposed

### LINKED ISSUE #406: Implement the rollup merklization utilities
In order to create a rollup chain we will need to ingest some number of transitions and then merklize them. This way we can store just the root hash. Additionally, we will need to prove inclusion once we’ve rolled it up. To do this we’ll need a merklization library!

This library should allow:
1. Computing the root of a tree after having passed in some number of leaf nodes.
2. Verify inclusion against the root.

This will use a SMT probably

karlfloersch:
## Description
This PR adds merkle tree utilities which will be used for the rollup chain. This includes: 

1. generating a merkle tree root;
2. verifying inclusion of elements in a merkle tree.

Note 1: this is *not* the sparse merkle tree implementation and is instead a normal merkle tree to be used for rolling up transactions. These tests also make use of a reasonable merkle tree library & so didn’t have to implement the merkle tree client, thankfully.

Note 2: I deleted a *lot* of unused contract code. It was our old Plasma deposit contract. Unfortunately it makes this PR a bit messy & I considered breaking that out into it’s own PR, but for now I’ll keep it here.

## Questions
- None yet

### Fixes
- Fixes #406

## Contributing Agreement


karlfloersch:
Closing as this was included in #448

PR #407, opened 2019-08-26
[WIP] RollupChain contract

### LINKED ISSUE #395: Implement `RollupChain` contract

karlfloersch:
## Description
When complete this PR will add a RollupChain contract which:

1. Can ingest an array of datablocks & package them into a rollup block.
2. Has a `proveTransitionInvalid(...)` function which takes in two transitions which were rolled up and then checks to see if the 2nd transition is valid. If it is it halts the chain.

## Questions
- None right this moment that I can think of

### Fixes
- Fixes #395
- Fixes #
- Fixes #

### Blockers
- #406

## Contributing Agreement


karlfloersch:
Going to close this for now to reduce clutter and can reopen when we get back to the RollupChain contract. — currently the priorities are 1) SMTs and 2) high level end-to-end mock.

PR #428, opened 2019-09-04
Implement MerkleTree.getMerkleProof(...)

### LINKED ISSUE #411: Add getMerkleProof(...) to MerkleTree
Right now it's suggested in the MerkleTree interface, but it will be required for other tasks in this PoC. Add Merkle proof calculation to our MerkleTree.

Note: Merkle proof calculation will need to acquire the treeLock in OptimizedSparseMerkleTree.

willmeister:
## Description
* Adds `getMerkleProof(...)` on `MerkleTree` interface and implements it for `OptimizedSparseMerkleTree`
* Adds unit tests for `getMerkleProof(...)`
* Fixes bug where proof siblings stored within `verifyAndStore(...)` are stored with the wrong key (actual key instead of sibling key)

### Fixes
- Fixes #411 

## Contributing Agreement

### DIFF
--- a/packages/core/src/app/utils/merkle-tree.ts
+++ b/packages/core/src/app/utils/merkle-tree.ts
@@ -9,6 +9,7 @@ import {
   HashFunction,
   MerkleTreeInclusionProof,
   MerkleTreeNode,
+  ONE,
   SparseMerkleTree,
   TWO,
   ZERO,
@@ -161,17 +162,17 @@ export class OptimizedSparseMerkleTree implements SparseMerkleTree {
     })
   }
 
-  public async update(key: BigNumber, value: Buffer): Promise<boolean> {
+  public async update(leafKey: BigNumber, leafValue: Buffer): Promise<boolean> {
     return this.treeMutex.runExclusive(async () => {
-      const nodesToUpdate: MerkleTreeNode[] = await this.getNodesInPath(key)
+      const nodesToUpdate: MerkleTreeNode[] = await this.getNodesInPath(leafKey)
       if (!nodesToUpdate) {
         return false
       }
 
       const leaf: MerkleTreeNode = nodesToUpdate[nodesToUpdate.length - 1]
       const idsToDelete: Buffer[] = [this.getNodeID(leaf)]
-      leaf.hash = this.hashFunction(value)
-      leaf.value = value
+      leaf.hash = this.hashFunction(leafValue)
+      leaf.value = leafValue
 
       let updatedChild: MerkleTreeNode = leaf
       let depth: number = nodesToUpdate.length - 2 // -2 because this array also contains the leaf
@@ -179,8 +180,8 @@ export class OptimizedSparseMerkleTree implements SparseMerkleTree {
       // If we're not updating all nodes, there's a shortcut node to update.
       if (this.height !== nodesToUpdate.length) {
         const ancestorHash: Buffer = this.getZeroHashAncestorFromLeaf(
-          key,
-          value,
+          leafKey,
+          leafValue,
           this.height - (nodesToUpdate.length - 1)
         )
 
@@ -190,8 +191,8 @@ export class OptimizedSparseMerkleTree implements SparseMerkleTree {
         updatedChild.hash = ancestorHash
         updatedChild.value = this.createLeafShortcutNode(
           ancestorHash,
-          value,
-          key,
+          leafValue,
+          leafKey,
           depth--
         ).value
       }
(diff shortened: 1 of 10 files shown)


willmeister:
> Great stuff! 😃 Only thing I noticed that currently, the `leafValue` passed for `getMerkleProof` is never actually used other than in the returned object itself, so it seems unnecessary. One option would be to authenticate that it is indeed the value at the requested leafKey, alternatively just removing the argument and using `getLeaf` internally is an option. What do you think?

That's a good call. I added a check because that's something we should definitely check. Also changed Optimized SMT implementation to be non-optimized to make things easier.

packages/core/src/app/utils/merkle-tree.ts
```diff
@@ -9,30 +9,19 @@ import {
...
- *
- * To indicate that a node's value is a pointer to a leaf node instead of the typical
- * value of (leftChildHash + rightChildHash), the value is stored as 65 bytes instead
- * of the typical 64 bytes, with the first byte being a dummy byte.
+ * SparseMerkleTree implementation assuming a 256-bit hash algorithm is used.
  */
-export class OptimizedSparseMerkleTree implements SparseMerkleTree {
+export class SparseMerkleTreeImpl implements SparseMerkleTree {
   public static readonly emptyBuffer: Buffer = new Buffer(32).fill('\x00')
   private static readonly siblingBuffer: Buffer = new Buffer(1).fill('\x00')
```
karlfloersch: I think that these `new Buffer(...)`s can be replaced with `Buffer.alloc(32)` which will fill with zeros by default & is the new suggested way to allocate new buffers

packages/core/src/app/utils/merkle-tree.ts
```diff
@@ -161,48 +126,28 @@ export class OptimizedSparseMerkleTree implements SparseMerkleTree {
     })
   }
 
-  public async update(key: BigNumber, value: Buffer): Promise<boolean> {
+  public async update(leafKey: BigNumber, leafValue: Buffer): Promise<boolean> {
```
karlfloersch: I already mentioned this in our call but of course long term we'll probably want 1) to use native `bigint` in the backend of our BigNumber library / replace with bitint directly / replace with buffer directly.

That said, none of these are worth the effort pre-devocon.

packages/core/test/app/utils/merkle-tree.spec.ts
```diff
@@ -456,4 +472,180 @@ describe('OptimizedSparseMerkleTree', () => {
...
+      hash = zeroHash
+      assert(proof.siblings[1].equals(hash))
+      hash = hashFunction(
+        hashBuffer.fill(hashFunction(zeroData), 0, 32).fill(zeroHash, 32)
+      )
+      assert(proof.siblings[0].equals(hash))
+
+      assert(proof.rootHash.equals(await tree.getRootHash()))
+    })
+  })
```
karlfloersch: Went through the tests and they look good! I did write a little benchmark test in addition here:

```
    it.only('can make a whole lot of updates at random places in the tree', async () => {
      const tree: SparseMerkleTree = new SparseMerkleTreeImpl(
        db,
        undefined,
        24,
        hashFunction
      )

      const startTime = +new Date()
      for (let i = 0; i < 100; i++) {
        const key = new BigNumber(Math.floor(Math.random()*50000))
        await tree.update(key, Buffer.from('yo what is gucci'))
      }
      const finishTime = +new Date()
      const durationInMiliseconds = finishTime - startTime
      console.log('Duration:', durationInMiliseconds)
    }).timeout(9000)
```

Could consider adding that at the bottom but commented out, or just not at all because this benchmark is truly a hasty and ugly construction lol :P

karlfloersch:
Oh another note that I totally forgot -- you could consider moving the merkle tree into the `block-production` directory / rename the `block-production` directory to `merklization` or something. Just because we've got other merklization code in the block-production section

packages/core/src/app/utils/merkle-tree.ts
willmeister: Done 👍

packages/core/test/app/utils/merkle-tree.spec.ts
willmeister: I added them and have them uncommented. They go very quickly, so I'm not too worried about having them in.

willmeister:
> Oh another note that I totally forgot -- you could consider moving the merkle tree into the `block-production` directory / rename the `block-production` directory to `merklization` or something. Just because we've got other merklization code in the block-production section

Sounds good. Moved it to `block-production` 👍

PR #429, opened 2019-09-04
Client-side executeTransaction(...)

### LINKED ISSUE #387: Implement client-side `executeTransaction`
A pure function which runs some `txData` on some accounts/`inputs` according to our smart contracting rules. Accepts an array of `inputs` based on the transaction, and `data`--the data included at the `input`th position in the tree. Also accepts some arbitrary `txData` e.g. a signature.

Returns an array of `outputData[]` corresponding to the new state at each `input`.

```typescript=
// transaction accounts being used in a transaction.
type txInput = {
    account: address,
    data: any
}
// function being used to execute the transaction
function executeTransaction(
    inputs: txInput[],
    txData: any
): bytes[]
```

The behavior of this function should directly map to its on-chain equivalent, which should be built by the time this is being tackled.

Note // this could be implemented as a wrapper around some evm-js implementation. Possibly Kelvin did some work on this in the past.  For this MVP though, the functions should be pretty simple, so probably fine to simulate/rewrite in JS.

willmeister:
## Description
Really this just builds on Karl's work and adds exchange fees to swaps

### Fixes
- Fixes #387 

## Contributing Agreement


karlfloersch:
Will do a full review soon but just wanted to quickly note that I think you `git add`ed an unneeded package-lock.json which is making this PR deceptively large 😁.

willmeister:
> Will do a full review soon but just wanted to quickly note that I think you `git add`ed an unneeded package-lock.json which is making this PR deceptively large 😁.

@karlfloersch Are you saying that I updated it when I shouldn't have or that this was added and should not be added. If the former, I can revert. If the latter, any reason why we wouldn't version it? NPM seems to indicate that it should be versioned: https://docs.npmjs.com/files/package-lock.json

packages/wallet/src/mock-rollup-client.ts
```diff
@@ -28,7 +28,7 @@ export class MockRollupClient {
    *             TODO: replace sign(...) with a reference to the keystore.
    */
   constructor(
-    readonly db: KeyValueStore,
+    private readonly db: KeyValueStore,
```
karlfloersch: Thank you!

packages/wallet/src/mock-rollup-state-machine.ts
```diff
@@ -56,22 +56,24 @@ export class InvalidTransactionTypeError extends Error {
...
   public getBalances(account: Address): Balances {
     if (!(account in this.state)) {
-      return {
-        uni: 0,
-        pigi: 0,
+      this.state[account] = {
+        balances: {
+          uni: 0,
+          pigi: 0,
+        },
```
karlfloersch: Ah meant to comment on this in my PR review, but one thing that we don't want to do is store anything in the state based on a `getBalances()` query. Basically because that means that someone could technically bloat our state by simply asking for random balances they know are zero. That's why I simply returned the value

packages/wallet/src/mock-rollup-state-machine.ts
```diff
@@ -99,19 +101,9 @@ export class MockRollupStateMachine {
     }
   }
 
-  private hasEnoughMoney(
-    account: string,
-    tokenType: TokenType,
-    greaterThanAmount: number
-  ) {
+  private hasBalance(account: Address, tokenType: TokenType, balance: number) {
```
karlfloersch: 😁 I like the name

packages/wallet/src/mock-rollup-state-machine.ts
willmeister: Ahh ok. I can protect it from cases like that. Will update 👍

willmeister: Ok, this should be fixed now.

karlfloersch: Woot!

PR #431, opened 2019-09-05
No need to verifyAndStore(...) empty leaves before update(...)

### LINKED ISSUE #430: Make updates for empty leaves not require calling verifyAndStore(...) beforehand

willmeister:
## Description
Makes it so that `update(...)` can be called without `verifyAndStore(...)` for empty leaf updates.

### Fixes
- Fixes #430 

### Blockers
- #428 Since this has 428's changes merged in

## Contributing Agreement


packages/core/src/app/block-production/merkle-tree.ts
```diff
@@ -127,11 +127,36 @@ export class SparseMerkleTreeImpl implements SparseMerkleTree {
...
+          nodesToUpdate.length
+        ))
+      ) {
         return false
       }
+      nodesToUpdate = await this.getNodesInPath(leafKey)
+    }
+
+    return this.treeMutex.runExclusive(async () => {
+      // Have to check to make sure nodesToUpdate didn't change while we didn't have the lock.
```
karlfloersch: I am probably missing some stuff but why not keep a lock the entire time that you're doing the update? Just to reduce a little complexity in this code. Is that possible?

packages/core/test/app/block-production/merkle-tree.spec.ts
```diff
@@ -684,16 +658,12 @@ describe('OptimizedSparseMerkleTree', () => {
...
 
-    it.only('updates: 100, treeHeight: 160, keyRange: 50,000', async () => {
+    it('updates: 100, treeHeight: 160, keyRange: 50,000', async () => {
       await runUpdateTest(160, 100, 50_000)
     }).timeout(9000)
-
-    it.only('updates: 1000, treeHeight: 160, keyRange: 50,000', async () => {
-      await runUpdateTest(160, 1000, 50_000)
-    }).timeout(9000)
   })
```
karlfloersch: Nice these tests look way simpler with update!

packages/core/src/app/block-production/merkle-tree.ts
willmeister: There's another call that locks, so this would deadlock if the lock is held. Other languages have the concept of re-entrant locks where if your thread is the one holding the lock, you can access anything protected by a lock, but unfortunately the mutex library doesn't do this, and my limited research shows that there's no easy swap-in replacement for it that does (since JS is single-threaded, and it may be hard to keep track of parallel contexts).

willmeister: That said, I agree with you and am not happy with how ugly this looks.

karlfloersch: Ok that's fair. Then well it sounds like you did the due diligence here & so imo let's just roll with the ugly code & clean it up later as at least this technical debt is isolated.

PR #442, opened 2019-09-12
Feat/425/get balances receipt

### LINKED ISSUE #425: State Querying and Receipt Generation
- accept getBalance from rest api
- use state to get inclusion proof of relevant state in current head state
- sign current head state
- return signature, head state, inclusion proof

willmeister:
## Description
Adds signatures to receipts for balances as well as roots and proofs

### Fixes
- Fixes #425 

## Contributing Agreement


packages/wallet/src/types/types.ts
```diff
@@ -61,3 +67,57 @@ export interface SignatureProvider {
...
+  updatedStateInclusionProof: StateInclusionProof
+}
+
+export interface SignedTransactionReceipt {
+  transactionReceipt: TransactionReceipt
+  signature: Signature
+}
+
+export interface StateReceipt {
+  address: string
```
ben-chain: Just to confirm, will this `address` always be the same as the key to the `state` below, right?  Is it just there to make looking up easier?

On a separate note--I think we'll want a transition/block number here being signed too, so that the aggregator attestations are easier to penalize for being faulty--without them, it's a little trickier to handle.

willmeister: Yep easy lookup and as a way to say "I'm guaranteeing _this_ state is as follows".

ben-chain:
Two comments:

1) To add on the note above that the state receipt should include block and transition numbers--perhaps this would be reason to actually query state receipts based on the pending block.  That would have the additional nice property of not having to lock the state from being updated when we query the state, since it's being pulled instead from the pending block store which will be append-only.

2) I see that you added a distinct `FaucetRequest` transaction type--I'm a little confused as to whether this should be an actual transaction type, since the faucet operation is just a send.  We do need to authenticate the faucet sends via the twitter API microservice, but to me that feels "out of protocol" in that it won't be reflected in anything on-chain.  Ping @karlfloersch who is actually doing the contracts--thoughts?

ben-chain:
Oops--just realized that my suggestion in 1) doesn't work, because the pending block has roots but we don't have the ability to serve inclusion proofs from it.  Never mind 😅

packages/wallet/src/types/types.ts
willmeister: Also, added block number and transition index to StateReceipt 👍

willmeister:
> Two comments:
> 
> [EDIT: ignore this one, doesn't work] 1) To add on the note above that the state receipt should include block and transition numbers--perhaps this would be reason to actually query state receipts based on the pending block. That would have the additional nice property of not having to lock the state from being updated when we query the state, since it's being pulled instead from the pending block store which will be append-only.
> 
> 1. I see that you added a distinct `FaucetRequest` transaction type--I'm a little confused as to whether this should be an actual transaction type, since the faucet operation is just a send.  We do need to authenticate the faucet sends via the twitter API microservice, but to me that feels "out of protocol" in that it won't be reflected in anything on-chain.  Ping @karlfloersch who is actually doing the contracts--thoughts?

We might be able to make `FaucetRequest` just a send, but it's really multiple sends, and more importantly it's a send from an address other than the signer to the address that is the signer. I'm open to doing whatever, but do we specifically gain something by making it just a send?

ben-chain:
I agree it's not a big deal either way, but the main thing is consistency between the contract state machine and aggregator persisted state machine.

> importantly it's a send from an address other than the signer to the address that is the signer

I'm not quite sure what you mean--are you saying the user receiving a faucet fund is doing the signing?  I think it's just going to be the twitter hook which authenticates the faucet funding.

PR #444, opened 2019-09-16
[WIP] Add Rollup Contracts

karlfloersch:
## Description
This PR is a little too large.

It includes:
- EvaluateTransition
- RollupChain

## Questions
- TODO

### Fixes
TODO
- Fixes #
- Fixes #
- Fixes #

## Contributing Agreement

- [x] I have read and understood the [Plasma Group Contributing Guide and Code of Conduct](https://github.com/plasma-group/pigi/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #448, opened 2019-09-18
Rollup Contracts

### LINKED ISSUE #392: Implement SMT contract with partial proofs
We must be able to, in Solidity, supply a root of a SMT and then fill in some number (not all) of the nodes which are contained in that tree. This will allow us to insert all of the “input” leafs which are touched in the transaction.

Additionally, we can keep this tree in storage so that when we’ve applied the state transition we can insert all new states into the tree & later compare the root to the state root that we expect. Note: this issue does not include insertion.

### LINKED ISSUE #393: Implement insertion in SMT contract
The next & hopefully last step in the on-chain SMT is implementing insertion. The insertion operations should be relatively easy as we are not rebalancing the tree. Note the insertions should be on the “input tree” which we earlier applied a partial proof to.

At the end we should be able to call `getRoot()` and get the root to compare with the committed root.

### LINKED ISSUE #406: Implement the rollup merklization utilities

karlfloersch:
## Description
This PR adds sparse merkle trees to the rollup contracts. It builds heavily on https://github.com/plasma-group/pigi/pull/444 but I figured I might as well split it up a bit. 

This PR is making sure that we can

1. Generate SMT roots [COMPLETE]
2. VerifyAndStore partial proofs [not yet...]

This will be enough for the contracts & so hopefully wont be too difficult.

## Questions
- None yet.

### Fixes
- Fixes #392
- Fixes #393
- Fixes #406

Turns out these issues were pretty poorly defined so even though this PR isn't huge it should address all of these.

## Contributing Agreement


packages/contracts/contracts/RollupMerkleUtils.sol
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /* Methods */
+    /**
+     * @notice Set default hashes
+     */
+    function setDefaultHashes() private {
+        // Set the initial default hash.
+        // TODO: Replace this hardcoded bytes value with keccak256(0x000...000) or similar.
+        defaultHashes[0] = 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563;
```
willmeister: I  know from my tree stuff that this is the result of `keccack256(ZERO_BYTES32)` 😄 . Anyone  reason why this isn't calculated here, though?

packages/contracts/contracts/RollupMerkleUtils.sol
```diff
@@ -0,0 +1,268 @@
...
+     * @param _dataBlock The data block we're storing.
+     * @param _path The path from the leaf to the root / the index of the leaf.
+     * @param _siblings The sibling nodes along the way.
+     */
+    function store(bytes memory _dataBlock, uint _path, bytes32[] memory _siblings) public {
+        // First compute the leaf node
+        bytes32 computedNode = keccak256(_dataBlock);
+        for (uint i = 0; i < _siblings.length; i++) {
+            bytes32 parent;
+            bytes32 sibling = _siblings[i];
```
willmeister: *Important* I think our siblings are in different orders. For me the last sibling is the leaf sibling.
```
siblings = [
root_child_sibling
...
leaf_sibling
]
```

 If this makes the most sense, I'll update. Let me know.

packages/contracts/contracts/RollupMerkleUtils.sol
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /**
+     * @notice get the n'th bit in a uint.
+     *         For instance, if exampleUint=binary(11), getNth(exampleUint, 0) == 1, getNth(2, 1) == 1
+     * @param _intVal The uint we are extracting a bit out of
+     * @param _index The index of the bit we want to extract
+     * @return The bit (1 or 0) in a uint8
+     */
+    function getNthBitFromRight(uint _intVal, uint _index) public pure returns (uint8) {
```
willmeister: I think this can be `function isRightSibling(...) public pure returns (bool)` because it's not used in any other way.

packages/contracts/contracts/RollupMerkleUtils.sol
```diff
@@ -0,0 +1,268 @@
...
+        return uint8(_intVal >> _index & 1);
+    }
+
+    /**
+     * @notice Get the children of some parent in the tree
+     * @param _parent The parent node
+     * @return (rightChild, leftChild) -- the two children of the parent
+     */
+    function getChildren(bytes32 _parent) public view returns(bytes32, bytes32) {
+        return (tree.nodes[getLeftSiblingKey(_parent)], tree.nodes[getRightSiblingKey(_parent)]);
```
willmeister: Shouldn't these be `getLeftChildKey(...)` and `getRightChildKey(...)`?

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    function verifyEcdsaSignature(bytes memory _signature, bytes32 _hash, address _pubkey) private pure returns(bool) {
+        return true;
+    }
+
+    /**
+     * Return the tx type inferred by the length of bytes
+     */
+    function inferTransitionType(
```
willmeister: Goes without saying, but we'll need to be darn sure of these lengths or run the risk of this contract locking funds.

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+        bytes32 stateRoot;
+        uint32[2] memory storageSlots;
+        uint transitionType = inferTransitionType(rawTransition);
+        if (transitionType == CREATE_AND_TRANSFER_TYPE) {
+            dt.CreateAndTransferTransition memory transition = decodeCreateAndTransferTransition(rawTransition);
+            stateRoot = transition.stateRoot;
+            storageSlots[0] = transition.senderSlot;
+            storageSlots[1] = transition.recipientSlot;
+        }
+        if (transitionType == TRANSFER_TYPE) {
```
willmeister: else if

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+            dt.TransferTransition memory transition = decodeTransferTransition(rawTransition);
+            stateRoot = transition.stateRoot;
+            storageSlots[0] = transition.senderSlot;
+            storageSlots[1] = transition.recipientSlot;
+        }
+        if (transitionType == SWAP_TYPE) {
```
willmeister: else if

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+        }
+        return (stateRoot, storageSlots);
+    }
+
+    function getTransferTxHash(dt.TransferTx memory _transferTx) internal pure returns(bytes32) {
+        return keccak256(abi.encode(_transferTx.sender, _transferTx.recipient, _transferTx.tokenType, _transferTx.amount));
+    }
+
+    function verifyEmptyStorage(dt.Storage memory _storage) internal pure {
+        require(_storage.pubkey == 0x0000000000000000000000000000000000000000, "Pubkey of storage slot must be zero");
```
willmeister: Would make a `NULL_ADDRESS` constant

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+        require(_storage.balances[0] == 0, "Uni balance must be zero");
```
willmeister: Would make `UNI_INDEX` and `PIGI_INDEX` constants

packages/contracts/contracts/UnipigTransitionEvaluator.sol
```diff
@@ -0,0 +1,331 @@
...
+     * Get the hash of the storage value.
+     */
+    function getStorageHash(dt.Storage memory _storage) public pure returns(bytes32) {
+        // Here we don't use `abi.encode([struct])` because it's not clear
+        // how to generate that encoding client-side.
+        return keccak256(abi.encode(_storage.pubkey, _storage.balances[0], _storage.balances[1]));
+    }
+
+    /************
+     * Decoding *
```
willmeister: *Important* `tokenType` was (and still is) listed as a `bool` in all our client encoding/decoding, whereas here it's listed as `uint`.

packages/contracts/contracts/RollupChain.sol
```diff
@@ -0,0 +1,271 @@
...
+    /* Methods */
+    function isHalted() public view returns(bool) {
+        return halted;
+    }
+
+    function pruneBlocksAfter(uint blockNumber) internal {
+        for (uint i = blockNumber; i < blocks.length; i++) {
+            delete blocks[i];
+        }
+        halted = true;
```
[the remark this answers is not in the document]
karlfloersch: I just removed the halting variable because there's really no need to halt the chain in the first place

packages/contracts/contracts/RollupMerkleUtils.sol
```diff
@@ -0,0 +1,268 @@
...
+/* Internal Imports */
+import {DataTypes as dt} from "./DataTypes.sol";
+
+/*
+ * Merkle Tree Utilities for Rollup
+ */
+contract RollupMerkleUtils {
+    /* Structs */
+    // A partial merkle tree which can be updated with new nodes, recomputing the root
+    struct SparseMerkleTree {
```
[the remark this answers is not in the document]
karlfloersch: Deleted SparseMerkleTreeLib! That was some old artifact

packages/contracts/contracts/UnipigTransitionEvaluator.sol
karlfloersch: Oh so as it turns out, bool encoding is actually identical to `uint`, so that's not a big problem. That said, I saw this is changed in the new encoding which this uses now so hopefully not an issue anymore. 

One note somewhat related to this is that I am currently doing a very dumb manual conversion from `balances` being an object with keys of '0', & '1', into `balances` being an array where the first item is uni and the 2nd is pigi. You can see it here: https://github.com/plasma-group/pigi/pull/448/files#diff-364498878b2bb730b65a6a87f059e45aR475

This is pretty annoying unfortunately.

packages/contracts/contracts/RollupChain.sol
```diff
@@ -0,0 +1,271 @@
...
+        bytes32 postStateRoot;
+        uint32[2] memory preStateStorageSlots;
+        uint32[2] memory storageSlots;
+        // First decode the prestate root
+        (success, returnData) =
+            address(transitionEvaluator).call(
+                abi.encodeWithSelector(transitionEvaluator.getTransitionStateRootAndAccessList.selector, _preStateTransition)
+            );
+        // Emit the output as an event
+        emit DecodedTransition(success, returnData);
```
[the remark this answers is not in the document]
karlfloersch: Whoops! Great point

packages/contracts/contracts/UnipigTransitionEvaluator.sol
willmeister: Will this require a change on the client? It should be easy to update [de-]serialization so that we sign it the way you want it but also represent it in JS how we want.

karlfloersch: This is a weird thing where the way you *have* to pass it in to ethers.js is an array. This is because they do the struct serialization on their end and so they need it in an object which is represented the same way that it is serialized. Additionally, you can't have mappings in structs, so there's no way to represent normal objects. This means that we've got to do this somewhat ugly transformation, or on the client side also use an array.

karlfloersch:
yolo

PR #450, opened 2019-09-18
State  Receipt Ingestion in RollupOVM

### LINKED ISSUE #418: Transaction Receipt Ingestion to OVM
- send agg sig to message ingester
- send inclusion proof to message ingester
- track most recent state root based on agg. signature on rollup block number and transaction index

YOLO:
- in the future, add this to a guard / watch list to compare to state root on ethereum at future point in time

willmeister:
## Description
- Adds `MerkleInclusionProofDecider` to the OVM for checking the validity of merkle proofs
- Adds RollupOVM for handling provability checks and proofs

### Fixes
- Fixes #418 

## Contributing Agreement


packages/wallet/src/rollup-ovm.ts
```diff
@@ -0,0 +1,116 @@
...
+   * @param stateReceipt
+   * @param signer
+   */
+  public async isStateReceiptProvablyValid(
+    stateReceipt: StateReceipt,
+    signer: Address
+  ): Promise<boolean> {
+    // TODO: should we return the decision here and make the caller handle disputes,
+    //       or should this handle disputes?
+    //       I'm leaning toward this handling disputes.
```
karlfloersch: Yes I like this handling disputes as well

packages/wallet/src/rollup-ovm.ts
```diff
@@ -0,0 +1,116 @@
...
+      stateReceipt,
+      signer
+    )
+
+    return decision.outcome ? decision.justification : undefined
+  }
+
+  private async decideIfStateReceiptIsValid(
+    stateReceipt: StateReceipt,
+    signer: Address
```
karlfloersch: This could be aggregator

packages/wallet/src/rollup-ovm.ts
willmeister: Agreed. for now there's a public `getFraudProof(...)` function for getting a fraud  proof. When we hook this bad boy up to mainnet, we'll hook into that to get proof for a dispute and make it private if not used externally.

packages/wallet/src/unipig-wallet.ts
```diff
@@ -7,52 +7,191 @@ import {
...
+      )
+    } catch (e) {
+      logError(
+        log,
+        `An error persisting transaction receipt for Tx: ${serializeObject(
+          transaction
+        )}.`,
+        e
+      )
+    }
```
karlfloersch: So I had some conversations with Ben about how to structure the wallet / client library. Would love to hear your input on this. That said, here's what we came up with:

1. The overall package which people install is `ovm` -- eg. `npm install ovm` (I'm in communication with npm about giving us control of that package name & will hopefully get it soon).
2. The ovm contains a keystore (aka wallet) where people hold their keys.
3. The ovm contains a "transitioner" 
4. The ovm contains a "state solver" -- this has all of the decider logic.

## ovm.transitioner
- send
- swap
- etc

This gets user intent and translates it into L2 actions (rollup being the way we translate for now). It also handles signing

## ovm.stateSolver
This contains 3 components:
1. Computability model
2. Game model
3. State enforcer 

The computability model & game model is what we've been building with the deciders. The 'computability model' is a model of what can and cannot be computed by other agents in our system (aka no one can compute our signature but us). The 'game model' is our model of rollup & of course in our case will be super simplistic.

The state enforcer is what sends fraud proofs to the mainchain. This also will be super simplistic but it can live in it's own object maybe to signify how important & distinct it's role is.

## It's a state transition system!
https://en.wikipedia.org/wiki/Transition_system

We've got state & transitions which are derived based on agent modeling / information hiding! Very cool!

## What this means
Even though we don't have "full" implementations of all components, it's important that we have the structure communicated so that folks can look at it and understand it's relation to our full vision.

The only changes here would be I would suggest renaming this class to `transitioner` (feel free to think of other names, but this one is nice because it preserves the concept of facilitating a "state transition", and then maybe renaming what is currently called `ovm` to instead `state solver` and putting the state enforcer inside that.

I've attached a photo of the high level picture & will put this into a better diagram soon.

Hugs!

image

packages/example-rollup/index.ts
```diff
@@ -50,69 +56,71 @@ setTimeout(() => {
...
 
+const signatureDB: DB = newInMemoryDB()
+const signedByDB: SignedByDB = new SignedByDB(signatureDB)
+const signedByDecider: SignedByDecider = new SignedByDecider(
+  signedByDB,
+  Buffer.from(wallet.address)
+)
+const ovm: RollupOVM = new DefaultRollupOVM(signedByDB, signedByDecider)
+const rollupClient: RollupClient = new RollupClient(newInMemoryDB())
+const unipigWallet = new UnipigWallet(newInMemoryDB(), ovm, rollupClient)
```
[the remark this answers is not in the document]
willmeister: Can definitely create a factory function that creates a `UnipigWallet` with minimal effort, but the developer will certainly need to make a decision on how and where stuff will be stored.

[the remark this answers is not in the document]
willmeister: Created this easy constructor 👍

PR #459, opened 2019-09-24
Contract Compatibility

### LINKED ISSUE #457: Make Client Compatible with Contracts
This entails:
- Reversing the order of Siblings in MerkleProofs (first element should be the leaf sibling)
- Update `tokenType` to be `uint` instead of `bool`.

willmeister:
## Description
This entails ordering Merkle proofs the same way the smart contracts do (first element is leaf sibling) and updating `tokenType` to serialize as a `uint` instead of a `bool`.

### Fixes
- Fixes #457 

## Contributing Agreement


packages/core/src/app/block-production/merkle-tree.ts
```diff
@@ -32,16 +32,22 @@ export class SparseMerkleTreeImpl implements SparseMerkleTree {
   private readonly treeLock: AsyncLock = new AsyncLock({
     domainReentrant: true,
   })
+  private readonly hashFunction: (Buffer) => Buffer
   private readonly hashBuffer: Buffer = Buffer.alloc(64)
 
   constructor(
     private readonly db: DB,
     rootHash?: Buffer,
     private readonly height: number = 160,
```
karlfloersch: Is the tree height still kind of "off by 1" in that an inclusion proof for a depth 32 tree has 31 siblings? If so we should update it (doesn't have to be internally, but deff from the contract perspective) to be depth 32 has 32 siblings

willmeister: @karlfloersch I thought we agreed that the root counts in height. The root never has a sibling, so sibling count will always be depth - 1.

karlfloersch: yeah I was into the root counting as the height, but after using a bunch of merkle tree libraries it looks like it's super standard practice to say a tree of depth 1 has a root and some children.

if it's not terribly troubling do you think we can adjust to make the height == number of siblings? We can also leave as is tbh or I can change in the contracts, but seeing as it's non-standard i prefer either doing nothing for now or changing it here

willmeister: Just noting this for the record. As we discussed, yes, we should do this, but we'll wait until after Devcon. I'll created #472 to address it.

PR #460, opened 2019-09-24
[Reference] aggregator deployment fixes

karlfloersch:
## Description
This pull request includes a number of small fixes which were required to get the aggregator up and running.

One fix, which is hacky in this implementation but should be done properly, is starting to remove the dependency on the `DefaultWallet` and `WalletDB` classes. I thought they might be useful but the problem is they take *way too long* to encrypt and decrypt. This means you have to wait like 5 seconds in the browser.

Instead, I think we might as well just use ethers.js for all of our wallet needs. We can just pass in the wallet & call it a day.

## Questions
- None at the moment

## Contributing Agreement

### DIFF
--- a/packages/wallet/src/rollup-state-solver.ts
+++ b/packages/wallet/src/rollup-state-solver.ts
@@ -62,8 +62,10 @@ export class DefaultRollupStateSolver implements RollupStateSolver {
     stateReceipt: StateReceipt,
     signer: Address
   ): Promise<boolean> {
-    return (await this.decideIfStateReceiptIsValid(stateReceipt, signer))
-      .outcome
+    // TODO: Reenable the state root validity check
+    // return (await this.decideIfStateReceiptIsValid(stateReceipt, signer))
+    //   .outcome
+    return true
   }
 
   /**
@@ -101,10 +103,10 @@ export class DefaultRollupStateSolver implements RollupStateSolver {
           decider: this.merkleInclusionDecider,
           input: {
             merkleProof: {
-              rootHash: Buffer.from(stateReceipt.stateRoot),
+              rootHash: Buffer.from(stateReceipt.stateRoot, 'hex'),
               key: new BigNumber(stateReceipt.slotIndex),
               value: abiEncodeState(stateReceipt.state),
-              siblings: stateReceipt.inclusionProof.map((x) => Buffer.from(x)),
+              siblings: stateReceipt.inclusionProof.map((x) => Buffer.from(x, 'hex')),
             },
           },
         },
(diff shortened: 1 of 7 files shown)


packages/wallet/src/unipig-transitioner.ts
```diff
@@ -75,7 +77,8 @@ export class UnipigTransitioner extends DefaultWallet {
     stateSolver: RollupStateSolver,
     rollupClient: RollupClient,
     signatureVerifier: SignatureVerifier = DefaultSignatureVerifier.instance(),
-    signatureProvider?: SignatureProvider
+    signatureProvider?: SignatureProvider,
+    wallet?: ethers.Wallet
```
willmeister: I think we should add `getAddress(...)` to the `SignatureProvider` interface. I've run into this in other places as well. It makes sense that we'll need it pretty much everywhere we will use `SignatureProvider`.

packages/wallet/src/rollup-state-solver.ts
```diff
@@ -101,10 +103,12 @@ export class DefaultRollupStateSolver implements RollupStateSolver {
           decider: this.merkleInclusionDecider,
           input: {
             merkleProof: {
-              rootHash: Buffer.from(stateReceipt.stateRoot),
+              rootHash: Buffer.from(stateReceipt.stateRoot, 'hex'),
```
willmeister: I didn't realize `hexStrToBuf(...)` is a thing. We should probably use that everywhere. 

ref: https://github.com/plasma-group/pigi/pull/460/files#diff-3c6a9a75c3d9298f45efe6c6dab74ecdR388

packages/wallet/src/unipig-transitioner.ts
```diff
@@ -87,6 +90,26 @@ export class UnipigTransitioner extends DefaultWallet {
     this.db = db
     this.stateSolver = stateSolver
     this.knownState = {}
+    this.wallet = wallet
+  }
+
+  public async sign(signer: string, message: string): Promise<string> {
+    if (typeof this.wallet !== 'undefined') {
+      log.debug('Were signing with ethers!')
```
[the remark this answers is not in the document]
karlfloersch: Lol tbh I meant to remove this because really I was using it as a quick debugging test. That said, I updated the log message to be a bit more descriptive and kept it in. Feel free to delete though if you think it's too noisey.

packages/wallet/src/rollup-state-solver.ts
karlfloersch: Yeah I noticed going through the code that you might not have realized we had that

PR #470, opened 2019-09-25
Simplify Transitioner

### LINKED ISSUE #461: Simplify transitioner & wallet
The UnipigTransitioner & the DefaultWallet are a bit over complicated. Issues include:

1) UnipigTransitioner extends the wallet class but that doesn't seem right... instead maybe it should just be passed a reference.
2) The wallet class does very resource intensive encryption. Instead we should either have the ability to toggle encryption in our current wallet implementation, or a simplified wallet class.

willmeister:
## Description
- Simplifying UnipigTransitioner by making it accept a SignatureProvider instead of being a DefaultWallet
- Making UnipigTransitioner only act on behalf of one Wallet instead of potentially many
- Changing SignatureProvider interface to provide address access
- Changing DefaultWallet to not be a SignatureProvider but instead have many SignatureProviders
- Updating test

### Fixes
- Fixes #461 

## Contributing Agreement


packages/example-rollup/index.ts
```diff
@@ -105,7 +104,7 @@ async function fetchBalanceUpdate() {
 }
 
 async function onRequestFundsClicked() {
-  await unipigWallet.requestFaucetFunds(wallet.address, 10)
+  await unipigWallet.requestFaucetFunds(10)
```
karlfloersch: 👍 Makes sense to do it like this

willmeister:
> This looks great! I noticed that the diff is broken for some reason, but I filtered it and only looked through the last commit. The last commit cleaning up the transitioner looks great to me.
> 
> One concern is if somehow merging this messes something up because the diff is so broken, though this may not be an issue. Also it's possible that me only reviewing the last commit missed some stuff... that being said still I'll approve as generally this looks solid / straightforward!

I merged `master`, and this should be much cleaner now. Sorry for not doing it earlier.

PR #473, opened 2019-09-27
[WIP] Generic L1 Communication

### LINKED ISSUE #348: Create Generic L1 Communication Capabilities
Communicate with L1 (should be generic if possible to support different L1s).

At a minimum should support:
    - Subscription to L1 events
        - Events
        - Logs
        - Errors
        - etc.
Can use: https://github.com/plasma-group/pigi/tree/master/packages/watch-eth

Nice to have:
    - Transaction submission to L1
        - Should include:
            - Subscribing to information about submitted transactions
            - Signing, validation, etc.

willmeister:
## Description
Handles event subscription, block subscription, and ability to submit transactions to Ethereum.

### Fixes
- Fixes #348 

## Contributing Agreement


packages/contracts/package.json
```diff
@@ -41,7 +41,7 @@
     "mocha": "^6.0.2",
     "rimraf": "^2.6.3",
     "typescript": "^3.3.3333",
-    "web3": "^1.0.0-beta.48"
+    "web3": "^1.2.1"
```
karlfloersch: Wonder if we can remove web3 entirely from dependencies?

packages/watch-eth/test/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  public async handle(t: T): Promise<void> {
+    log.debug(`Received ${JSON.stringify(t)}`)
+    this.received.push(t)
+  }
+
+  public getReceived(): T[] {
+    return this.received.splice(0)
+  }
+
+  public async waitForReceive(num: number = 1): Promise<T[]> {
```
karlfloersch: Ah ok was confused about this at first, but to make sure I understand: this is a test helper which allows us to wait until we've received all of the messages which we expect & then returns. This will hang if we don't get all the messages but that's ok because the test will fail because of the timeout.

packages/watch-eth/test/ethereum-block-processor.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    await tokenContract.transfer(
+      ownerWallet.address,
+      recipientWallet.address,
+      sendAmount
+    )
+
+    const blocks: Block[] = await blockListener.waitForReceive(1)
+
+    blocks.length.should.equal(1)
```
karlfloersch: This can never be less than one, it can only be 1 or more (based on the `waitForReceive()` function) right?

packages/watch-eth/test/ethereum-event-processor.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+      recipientWallet.address,
+      sendAmount
+    )
+
+    const events = await eventListener.waitForReceive()
+    events.length.should.equal(1)
+    const event: Event = events[0]
+    event.values['from'].should.equal(ownerWallet.address)
+    event.values['to'].should.equal(recipientWallet.address)
+    event.values['amount'].toNumber().should.equal(sendAmount)
```
karlfloersch: Great

packages/watch-eth/test/ethereum-event-processor.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+  }).timeout(timeout)
+
+  it('processes old events', async () => {
```
karlfloersch: You said that this implementation also stores the events which have already been processed so if we want to we can skip through all of these events I presume?

packages/watch-eth/test/utils.ts
willmeister: Yep! 😄

packages/watch-eth/test/ethereum-block-processor.spec.ts
willmeister: Correct!

packages/watch-eth/test/ethereum-event-processor.spec.ts
willmeister: Yep, we store the last successfully processed block [here](https://github.com/plasma-group/pigi/pull/473/files#diff-4fdcbba76dbc4f9595f420a5666270bcR62) and use that as our `start` when we're backfilling things that we missed when we startup.

willmeister: Also can pass `false` in for the `backfill` parameter to `subscribe(...)` (defaults to `true`) or pass a start block number to the constructor.

karlfloersch: awesome!

PR #474, opened 2019-09-27
Handling Events in Aggregator

### LINKED ISSUE #452: Handle Aggregator Failure
This could be Aggregators always picking up exactly where they left off but more likely is the wallet acknowledging that the Aggregator's signature means nothing if they don't publish another block. Need to think through this UX and make sure we're covered.

In our case, we'll only have one aggregator (maybe represented by many different services), so that aggregator cannot make promises, fail, and then come up and start new, publishing new blocks without incorporating those promises.

### LINKED ISSUE #464: Submit aggregator blocks to Ethereum
Should be as simple as sending an ethereum transaction which contains all of the abi encoded transitions to:

https://github.com/karlfloersch/pigi/blob/feat/contract_merklization/packages/contracts/contracts/RollupChain.sol#L50-L63

This should probably get triggered every minute OR when we get to 300 transitions.

### LINKED ISSUE #465: Add aggregator block event watcher
Watch blocks that are submitted through ethereum events (they will be ingested). This will also encompass syncing the aggregator with previous blocks (confirms for blocks it submitted) if it went down.

### LINKED ISSUE #475: Add block submission (L1) queue
Add a queue which will accept blocks and only submit new L1 transactions when the previous block has been confirmed/mined.

willmeister:
## Description
Handling events in aggregator
- Created an AggregatorServer class to separate the  networking from the logic
- Created a BlockSubmitter class to handle block queueing and submission
- Made Aggregator pass new block events to block submitter
- Made Aggregator survive restarts and added re-initialization tests
- Made BlockSubmitter survive restarts 
- Added BlockSubmitter tests for all functionality and restarts

Also:
- Made Aggregator survive restarts and added tests for it

### Fixes
- Fixes #452 
- Fixes #464 
- Fixes #465 
- Fixes #475 

## Contributing Agreement

### DIFF
--- a/packages/wallet/src/default-rollup-block-submitter.ts
+++ b/packages/wallet/src/default-rollup-block-submitter.ts
@@ -72,6 +72,15 @@ export class DefaultRollupBlockSubmitter implements RollupBlockSubmitter {
   }
 
   public async submitBlock(rollupBlock: RollupBlock): Promise<void> {
+    if (rollupBlock.number <= this.lastQueued) {
+      log.error(
+        `submitBlock(...) called on old block. Last Queued: ${
+          this.lastQueued
+        }, received: ${JSON.stringify(rollupBlock)}`
+      )
+      return
+    }
+
     this.blockQueue.push(rollupBlock)
     await this.db.put(
       DefaultRollupBlockSubmitter.getBlockKey(rollupBlock.number),
--- a/packages/wallet/src/default-rollup-block-submitter.ts
+++ b/packages/wallet/src/default-rollup-block-submitter.ts
@@ -183,7 +183,7 @@ export class DefaultRollupBlockSubmitter implements RollupBlockSubmitter {
   public static serializeRollupBlockForSubmission(
     block: RollupBlock
   ): string[] {
-    return block.transitions.map((x) => x.stateRoot)
+    return block.transitions.map((x) => abiEncodeTransition(x))
   }
 
   public static serializeRollupBlockForStorage(
--- a/packages/wallet/src/default-rollup-block-submitter.ts
+++ b/packages/wallet/src/default-rollup-block-submitter.ts
@@ -2,15 +2,20 @@ import { DB, getLogger } from '@pigi/core'
 import { Block } from 'ethers/providers'
 import { Contract } from 'ethers'
 
-import { RollupBlock } from './types'
+import { RollupBlock, RollupBlockSubmitter } from './types'
+import { abiEncodeTransition, parseTransitionFromABI } from './serialization'
 
 const log = getLogger('rollup-block-submitter')
 
-const LAST_CONFIRMED_KEY: Buffer = Buffer.from('last_confirmed')
-const LAST_SUBMITTED_KEY: Buffer = Buffer.from('last_submitted')
-const LAST_QUEUED_KEY: Buffer = Buffer.from('last_queued')
+export class DefaultRollupBlockSubmitter implements RollupBlockSubmitter {
+  public static readonly LAST_CONFIRMED_KEY: Buffer = Buffer.from(
+    'last_confirmed'
+  )
+  public static readonly LAST_SUBMITTED_KEY: Buffer = Buffer.from(
+    'last_submitted'
+  )
+  public static readonly LAST_QUEUED_KEY: Buffer = Buffer.from('last_queued')
 
-export class DefaultRollupBlockSubmitter {
   private lastSubmitted: number
   private lastConfirmed: number
   private lastQueued: number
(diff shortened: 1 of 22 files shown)


packages/wallet/test/helpers.ts
```diff
@@ -106,14 +110,15 @@ export const calculateSwapWithFees = (
   )
 }
 
-export const assertThrows = (func: () => any, errorType: any): void => {
+export const assertThrows = (func: () => any, errorType?: any): void => {
```
ben-chain: Just curious if this has any different functionality from `should.throw()` :)

packages/wallet/test/rollup-block-submitter.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+        blockSubmitter,
+        dummyContract,
+        2,
+        2,
+        0,
+        [rollupBlock, rollupBlock2]
+      )
+    })
+
+    it('should try to submit when one queued but not submitted', async () => {
```
ben-chain: In theory, a more robust test would actually check that an attempted submission to L1 happens here, right?

packages/wallet/test/rollup-block-submitter.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+      const lastQueuedFromDB: number = await getLastQueuedFromDB(db)
+      lastQueuedFromDB.should.equal(2)
+    })
+  })
+
+  describe('handleNewRollupBlock()', () => {
+    it('should do nothing when there are no pending blocks', async () => {
+      await blockSubmitter.init()
+
+      await blockSubmitter.handleNewRollupBlock(1)
```
ben-chain: Still slightly confused by this function--why is `1` passed here?  Is this simulating a multi-aggregator world where block 1 was submitted by someone else?

packages/wallet/src/default-rollup-block-submitter.ts
```diff
@@ -0,0 +1,206 @@
...
+    this.lastQueued = rollupBlock.number
+    await this.db.put(
+      DefaultRollupBlockSubmitter.LAST_QUEUED_KEY,
+      this.numberToBuffer(this.lastQueued)
+    )
+
+    await this.trySubmitNextBlock()
+  }
+
+  public async handleNewRollupBlock(rollupBlockNumber: number): Promise<void> {
```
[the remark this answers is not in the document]
willmeister: It's when a rollup block event is emitted, which should contain a block number. There may be multiple rollup blocks in one L1 block, but probably not.

packages/wallet/src/default-rollup-block-submitter.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      // If we failed after submission but before storing submitted, update lastSubmitted
+      if (this.lastSubmitted < this.lastConfirmed) {
+        this.lastSubmitted = rollupBlockNumber
+        await this.db.put(
+          DefaultRollupBlockSubmitter.LAST_SUBMITTED_KEY,
+          this.numberToBuffer(this.lastSubmitted)
+        )
+      }
+      await this.trySubmitNextBlock()
```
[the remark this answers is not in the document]
willmeister: Yes, this is right. `this.lastConfirmed` must be `this.lastSubmitted + 1`. It can't be larger since we only submit one at a time.

packages/wallet/test/helpers.ts
willmeister: Probably not, but I think I created this when creating the async version because `should.throw()` can't handle that. Should be safe to remove this one, though.

packages/wallet/test/rollup-block-submitter.spec.ts
willmeister: It does: https://github.com/plasma-group/pigi/pull/474/files/088cbabddece5a2cba2640574e890a47943ede68#diff-b05b44c2d5e0451d7661ca8d857b1a5fR87, although due to separation of concerns, that part is mocked, but we do assert that it calls the contract's `submitBlock(...)` function with the appropriate data.

packages/wallet/test/rollup-block-submitter.spec.ts
willmeister: This is simulating getting a new block event from the contract that says "rollup block 1 was mined"

packages/wallet/src/default-rollup-block-submitter.ts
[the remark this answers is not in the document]
willmeister: It's confusing because this isn't wired up yet, but the RollupAggregator `implements EthereumListener<Event>`, where it will subscribe to RollupChain.sol's `NewRollupBlock(...)` event. Right now that just has all of the transitions, but that will be changed to have the block number (and eventually get rid of the transitions).

packages/wallet/src/default-rollup-block-submitter.ts
```diff
@@ -0,0 +1,230 @@
...
+    await this.db.put(
+      DefaultRollupBlockSubmitter.LAST_SUBMITTED_KEY,
+      this.numberToBuffer(this.lastSubmitted)
+    )
+  }
+
+  public static serializeRollupBlockForSubmission(
+    block: RollupBlock
+  ): string[] {
+    return block.transitions.map((x) => x.stateRoot)
```
ben-chain: Hmmmm, seems that Jing's wifi squashed a comment I had left here or at least I can't find it now--only change I was actually requesting with this original review was that this should be `(x) => abiEncodeTransition(x)`; contract needs the full transition to be rolled up.

willmeister: Good call on this! Updated 👍

PR #479, opened 2019-10-01
[WIP] Rollup Validator

### LINKED ISSUE #466: Translate raw block (transitions) into transactions which can be applied
We need to look up the pubkey of storage slots for the transitions that we are translating in order to get the actual transaction which is something we can apply to our state machine. This means we've got to write a little converter script.

This means we need to find and replace `storageSlot` with -> `pubkey`

ben-chain:
## Description
WIP, adds a basic validating node.  Currently supports the following:
- ingest `RollupBlock`s
- calculate expected root and compare to that of the transitions' postRoot
- generate a fraud proof

Currently does not support:
- edge case where the fraudulent transition is the first in the block (fraud proof requires prev block)
- Create and Transfers -- running into an error when generating the Merkle proof for an unset storage slot, which is needed to generate the fraud proof.  See skipped test in `rollup-state-validator.spec.ts`
- persistent validation
- L1 integration
## Questions
- What's the best way to persist?
- How do we want to display this loading bar?  CLI style?

### Fixes
- Fixes # 421
- Fixes #466

### Blockers
- L1 event querying

## Contributing Agreement


ben-chain:
Note the way this implements #466 is to convert using both a `StateSnapshot` and `Transition`, not just the transition.  However, I think this can't be avoided, since we need to get the pubkey from the storage slot which can't be determined from just the transition.

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+import { Transaction } from 'ethers/utils'
+import {
+  parseTransitionFromABI,
+  parseTransactionFromABI,
+} from './serialization'
+
+const log = getLogger('rollup-guard')
+export class DefaultRollupStateValidator implements RollupStateValidator {
+  public rollupMachine: DefaultRollupStateMachine
+  public currentPosition: RollupTransitionPosition = {
```
willmeister: Can make this private in favor of the getter on line 101. 

Note: that getter would let an external actor change the `currentPosition` from outside of this class, so exposing that is generally frowned upon unless it's absolutely required. Can return a copy (`return {...this.currentPosition}`) to get around that.

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (isSwapTransition(transition)) {
+      const swapperSnapshot: StateSnapshot = await this.rollupMachine.getSnapshotFromSlot(
+        transition.senderSlotIndex
+      )
+      const uniSnapshot: StateSnapshot = await this.rollupMachine.getState(
+        UNISWAP_ADDRESS
+      )
+      return [swapperSnapshot, uniSnapshot]
+    } else if (isCreateAndTransferTransition(transition)) {
+      const nextAccountKey: number = this.rollupMachine.getNextNewAccountSlot()
```
willmeister: This appears to be unused

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+      )
+      const recipientSnapshot: StateSnapshot = await this.rollupMachine.getSnapshotFromSlot(
+        transition.recipientSlotIndex
+      )
+      return [senderSnapshot, recipientSnapshot]
+    } else if (isTransferTransition(transition)) {
+      const senderSnapshot: StateSnapshot = await this.rollupMachine.getSnapshotFromSlot(
+        transition.senderSlotIndex
+      )
+      const recipientSnapshot: StateSnapshot = await this.rollupMachine.getSnapshotFromSlot(
```
willmeister: Can just set senderSlot and recipientSlot in the if / else if / else and then do the `this.rollupMachine.getSnapshotFromSlot(...)` calls and the return once at the end.

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+      const swapper: Address = snapshots[0].state.pubKey
+      const convertedTx: Swap = {
+        sender: swapper,
+        tokenType: transition.tokenType as UniTokenType | PigiTokenType,
+        inputAmount: transition.inputAmount,
+        minOutputAmount: transition.minOutputAmount,
+        timeout: transition.timeout,
+      }
+      return {
+        signature: transition.signature,
```
willmeister: Will this signature still be valid since it's a signature for a Transition?

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+  public async checkNextTransition(
+    nextTransition: RollupTransition
+  ): Promise<FraudCheckResult> {
+    let preppedFraudInputs: StateSnapshot[]
+    let generatedPostRoot: Buffer
+
+    const transitionPostRoot: Buffer = hexStrToBuf(nextTransition.stateRoot)
+
+    if (isCreateAndTransferTransition(nextTransition)) {
+      const slotIfSequential: number = await this.rollupMachine.getNextNewAccountSlot()
```
willmeister: Noted before, but should be `(await this.rollupMachine.getNextNewAccountSlot()) + 1`

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+    // let inputAsTransaction: SignedTransaction = await this.getTransactionFromTransition(nextTransition)
+    const inputAsTransaction: SignedTransaction = await this.getTransactionFromTransitionAndSnapshots(
+      nextTransition,
+      preppedFraudInputs
+    )
+    try {
+      await this.rollupMachine.applyTransaction(inputAsTransaction)
+      generatedPostRoot = await this.rollupMachine.getStateRoot()
+    } catch (error) {
+      if (isStateTransitionError(error)) {
```
willmeister: Should we log here too? Maybe in all of the cases, not just this `if`

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+          fraudPosition: this.currentPosition,
+          fraudInputs: preppedFraudInputs,
+          fraudTransition: nextTransition,
+        }
+      } else {
+        throw new LocalMachineError()
+      }
+    }
+
+    if (generatedPostRoot.equals(transitionPostRoot)) {
```
willmeister: Would log in the `if` and `else`

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+      this.currentPosition.blockNumber++
+      this.currentPosition.transitionIndex++
+      return 'NO_FRAUD'
```
willmeister: Would `undefined` also work? `undefined` has the benefit of being able to check fraud or not with a single `!` in an `if`.

packages/wallet/src/rollup-state-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    if (nextBlock.blockNumber !== currentPosition.blockNumber + 1) {
+      throw new ValidationOutOfOrderError()
+    }
+
+    for (const transition of nextBlock.transitions) {
+      const fraudCheck: FraudCheckResult = await this.checkNextTransition(
+        transition
+      )
+      if (fraudCheck !== 'NO_FRAUD') {
```
willmeister: Can change `checkNextTransition(...)` to `getTransitionFraudProof(...)`, and then `if (!!fraudProof)` means there was fraud and you won't have to cast it to `LocalFraudProof` below.

packages/wallet/src/rollup-state-validator.ts
ben-chain: Cool, I never knew why your code had that! 😂

packages/wallet/src/rollup-state-validator.ts
ben-chain: Yes, confirmed with Karl that the transition is converted to a transaction before sig is checked in contracts.

packages/wallet/src/rollup-state-validator.ts
ben-chain: I thought that making sure it "indeed found no fraud" as opposed to some bug which resulted in undefined would be useful--if this is an anti-pattern, I can change.

ben-chain: disregard--changing as this helps with 'getTransitionFraudProof' below

packages/wallet/src/rollup-state-machine.ts
```diff
@@ -351,6 +351,57 @@ export class DefaultRollupStateMachine implements RollupStateMachine {
...
+      }
+    )
+    return lockedRoot
+  }
+
+  public getNextNewAccountSlot(): number {
+    return this.lastOpenKey.toNumber() + 1
+  }
+
+  public async getSnapshotFromSlot(key: number): Promise<StateSnapshot> {
```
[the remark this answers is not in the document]
ben-chain: I had done it this way because we need to get the empty slots for create-and-transfers; as these do not have addresses I do think we need this.

PR #492, opened 2019-10-03
Wire up validator and contract

### LINKED ISSUE #491: Add Validator Subscriptions
Validator should be wired up to get event subscription data.

willmeister:
## Description
- Creates a Validator application that can be run
- Adds environment configuration
- Wires up the Validator with the Contract to subscribe to data

### Fixes
- Fixes #491

## Contributing Agreement


packages/wallet/src/validator/rollup-fraud-guard.ts
```diff
@@ -0,0 +1,162 @@
...
+        }. Received: ${JSON.stringify(block)}`
+      )
+      return
+    } else if (this.lastBlockValidated + 1 !== block.blockNumber) {
+      log.error(
+        `Received event with block number greater than expected! lastValidated: ${
+          this.lastBlockValidated
+        }. Received: ${JSON.stringify(block)}`
+      )
+      process.exit(1)
```
ben-chain: Is there any chance that Ethereum events will come in out of order?  I'm not super familiar with ethereum event logging, so may be a silly question, but thought I'd mention because this would cause issues if that were to happen.  (and if it can happen but is unlikely, seems low priority to solve)

packages/wallet/src/validator/rollup-fraud-guard.ts
```diff
@@ -0,0 +1,162 @@
...
+      const receipt: TransactionReceipt = await this.contract.proveTransitionInvalid(
+        ...proof
+      )
+      log.error(`Fraud proof submitted. Receipt: ${JSON.stringify(receipt)}`)
+    } catch (e) {
+      logError(log, 'Error submitting fraud proof!', e)
+      process.exit(1)
+    }
+
+    log.info('Congrats! You helped the good guys win. +2 points for you!')
```
ben-chain: Love it

packages/wallet/src/validator/rollup-state-validator.ts
```diff
@@ -177,33 +163,39 @@ export class DefaultRollupStateValidator implements RollupStateValidator {
...
       )
       this.currentPosition.transitionIndex++
       return undefined
     } else {
-      log.info(
-        'Ingested valid transition and postRoot disagreed with the aggregator claim--returning fraud'
+      log.error(
+        `Ingested valid transition and postRoot disagreed with the aggregator claim--returning fraud. Transition: ${JSON.stringify(
+          nextTransition
+        )}. Expected State Root: ${hexBufToStr(generatedPostRoot)}`
```
ben-chain: Great to see how you did logging here--I guess the rule of thumb to have is if you're logging something that happened, also log the relevant objects causing it to happen so you can immediately get insight when debugging.  Makes sense!

PR #493, opened 2019-10-04
Full signature support/compatibility for contract/client/aggregator

ben-chain:
## Description
Adds sig checking to the contracts, and makes the necessary changes for those to be compatible with the aggregator and client.

Specifically, note that for a few reasons this makes it so that the signatures are actually on hashes of the underlying data.  The ``DefaultSignatureProvider`` and ``DefaultSignatureVerifier`` both do this now, and hopefully nothing will break--all tests pass so I think we're good.

It took me a while to track down an associated bug which stemmed from ethers always signing the ASCII representation of strings, even if they start with ``0x`` like all of ours were.  So, the strings are now converted to buffers before being hashed and signed.  This broke regular signing and hashing for ``serializeObject``s, because this just uses JSON.stringify which is not a hex string, and this was being done for the faucet requests since they never need the ABI treatment.  So, I added a ``serializeObjectAsHexString`` and used that for receipts both client and server side.
## Questions
- 
-
-

### Fixes
- Fixes #
- Fixes #
- Fixes #

### Blockers
- #
- #
- #

## Contributing Agreement

- [X] I have read and understood the [Plasma Group Contributing Guide and Code of Conduct](https://github.com/plasma-group/pigi/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

### DIFF
--- a/packages/core/src/app/serialization/serializers.ts
+++ b/packages/core/src/app/serialization/serializers.ts
@@ -1,5 +1,6 @@
 import { AddressBalance, StateChannelMessage } from './examples'
 import { BigNumber, Message } from '../../types'
+import { hexStrToBuf } from '../utils'
 
 /**
  * Serializes the provided object to its canonical string representation.
@@ -20,6 +21,31 @@ export const serializeObject = (obj: {}): string => {
  */
 export const deserializeObject = (obj: string): {} => {
   return JSON.parse(obj)
+}
+
+
+/**
+ * Serializes the provided object to its canonical hex string representation.
+ *
+ * @param obj The object to serialize.
+ * @returns The serialized object as a string.
+ */
+export const serializeObjectAsHexString = (obj: {}): string => {
+  const stringified: string = JSON.stringify(obj)
+  const asHexString = '0x' + Buffer.from(stringified, 'utf-8').toString('hex')
+  return asHexString
+}
+
+/**
+ * Deserializes the provided hex string into its object representation.
+ * This assumes the string was serialized using the associated serializer.
+ *
+ * @param obj The string to deserialize.
+ * @returns The deserialized object.
+ */
+export const deserializeObjectAsHexString = (obj: string): {} => {
+  const asBuffer = hexStrToBuf(obj).slice(2)
+  return JSON.parse(asBuffer.toString('utf-8'))
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/src/app/serialization/serializers.ts
```diff
@@ -22,6 +23,30 @@ export const deserializeObject = (obj: string): {} => {
...
+
+/**
+ * Deserializes the provided hex string into its object representation.
+ * This assumes the string was serialized using the associated serializer.
+ *
+ * @param obj The string to deserialize.
+ * @returns The deserialized object.
+ */
+export const deserializeObjectAsHexString = (obj: string): {} => {
+  const asBuffer = hexStrToBuf(obj).slice(2)
```
willmeister: use `remove0x` instead so that in the event that it doesn't start with `0x`, we don't remove the first byte.

packages/core/src/app/serialization/serializers.ts
```diff
@@ -22,6 +23,30 @@ export const deserializeObject = (obj: string): {} => {
...
 
+/**
+ * Serializes the provided object to its canonical hex string representation.
+ *
+ * @param obj The object to serialize.
+ * @returns The serialized object as a string.
+ */
+export const serializeObjectAsHexString = (obj: {}): string => {
+  const stringified: string = JSON.stringify(obj)
+  const asHexString = '0x' + Buffer.from(stringified, 'utf-8').toString('hex')
```
willmeister: `utf-8` is the default, right? I don't think you need to specify it.

packages/core/src/app/serialization/serializers.ts
```diff
@@ -22,6 +23,30 @@ export const deserializeObject = (obj: string): {} => {
...
+/**
+ * Deserializes the provided hex string into its object representation.
+ * This assumes the string was serialized using the associated serializer.
+ *
+ * @param obj The string to deserialize.
+ * @returns The deserialized object.
+ */
+export const deserializeObjectAsHexString = (obj: string): {} => {
+  const asBuffer = hexStrToBuf(obj).slice(2)
+  return JSON.parse(asBuffer.toString('utf-8'))
```
willmeister: `utf-8` is the default, right? I don't think you need to specify it.

PR #494, opened 2019-10-05
Faucet requests only from uniswap

willmeister:
## Description
Note: Branches off of #492 -- that one should be merged first.

Makes it so that faucet requests are only valid from a configurable address injected into the aggregator. In the mock-aggregator, that is set from an environment variable.

### Fixes
- Fixes #496 

## Contributing Agreement

- [x] I have read and understood the [Plasma Group Contributing Guide and Code of Conduct](https://github.com/plasma-group/pigi/blob/master/.github/CONTRIBUTING.md) and am following those guidelines in this pull request.

PR #495, opened 2019-10-05
Hide aggregator mnemonic

### LINKED ISSUE #497: Hide Aggregator mnemonic
Aggregator mnemonic is out in the open right now, and the aggregator address is hard-coded. These should both be configurable, and mnemonic should be private.

willmeister:
## Description
Note: Branches off of #494  -- That one should be merged first

Hides the aggregator mnemonic (configurable through env variable) and makes the aggregator address configurable instead of a constant.

### Fixes
- Fixes #497 

## Contributing Agreement


packages/wallet/src/validator/monitor-and-validate.ts
```diff
@@ -15,19 +15,20 @@ config({ path: resolve(__dirname, `../../../.env`) })
...
 } from '../index'
 
 const log = getLogger('monitor-and-validate')
 
-const rollupContractAddress = process.env.ROLLUP_CONTRACT_ADDRESS
-const mnemonic = process.env.WALLET_MNEMONIC
-const jsonRpcUrl = process.env.JSON_RPC_URL
+const aggregatorAddress: Address = process.env.AGGREGATOR_ADDRESS
+const rollupContractAddress: Address = process.env.ROLLUP_CONTRACT_ADDRESS
+const mnemonic: string = process.env.WALLET_MNEMONIC
```
karlfloersch: So I guess our mnemonic can potentially disagree with the `aggregatorAddress`? 

Maybe it should check that they are the same with something like: `Wallet.fromMnemonic(mnemonic).address === aggregatorAddress`?

willmeister: The mnemonic is the validator's mnemonic, not the aggregator's. So if the validator finds fraud, they'll submit a tx with their wallet.

PR #501, opened 2019-10-05
Protect submit block

### LINKED ISSUE #500: Protect RollupChain.submitBlock(...) so only aggregator can call it

willmeister:
## Description
NOTE: Branches off of #499 -- Review that one first!

Only allows aggregator to call submit block

### Fixes
- Fixes #500 

## Contributing Agreement


packages/contracts/config/.env.example
```diff
@@ -11,6 +11,7 @@
 
 # REQUIRED
 DEPLOY_MNEMONIC='your mnemonic here'
+AGGREGATOR_ADDRESS='0x17FC6B6c35f4C4A1e92D9d878aC001EBbF1d8e2D'
```
ben-chain: Might want to specify whether these are required to match.  I'm assuming the AGGREGATOR_ADDRESS is the pubkey to the corresponding aggregator mnemonic, not the deploy mnemonic, but could be confusing to those trying to spin up.

packages/contracts/deploy/deploy-rollup-chain.ts
```diff
@@ -69,12 +69,15 @@ const deployContracts = async (wallet: Wallet): Promise<void> => {
...
 
+  const aggregatorAddress: string = process.env.AGGREGATOR_ADDRESS
+
   console.log('Deploying RollupChain...')
   const rollupChain = await deployContract(
     RollupChain,
     wallet,
     evaluatorContractAddress,
-    merkleUtilsCnontractAddress
+    merkleUtilsCnontractAddress,
```
ben-chain: typo

packages/contracts/config/.env.example
willmeister: Yep, your assumption is right, but I'll clarify

PR #542, opened 2019-10-31
Pigi Package Re-org

### LINKED ISSUE #521: Design & propose new pigi package structure

### LINKED ISSUE #523: Implement new Pigi package structure
Delete unused code in pigi

willmeister:
## Description
Re-organinzing pigi package structure to make core package much smaller (splitting it into multiple smaller packages), deleting unused code, and merging related code in separate packages.

### Fixes
- Fixes #521 
- Fixes #523 

## Contributing Agreement


karlfloersch:
Because the diff here isn't too helpful, I'll break my comments up into chunks:

### core-db
- Could consider removing the `utils/range.ts` file as we probably won't be working with ranges so extensively anymore...
- Or might just consider moving everything other than the BaseDB / Bucket functionality out of this package. Seems like some things here are less "db" focused--unlike our DB wrappers which are a very well scoped & useful library. 

### core-utils
- If `abi-range.ts` is not being used in other places we could consider removing it (though might be used in plasma)
- If we really wanted to we could move `transport` out of `core-utils` and into it's own package, though may not be worth the effort.

### unipig
- This is great! I really like how all of the code for the demo is in one place--I'm sure it would make it easier for others to understand our relatively complex code.

## Thoughts
I think that this structure is a major improvement over what we had! The only meta feedback I had was as we build out our full rollup implementation, it could make sense to make use of multiple packages--like one for aggregator, client, etc--even if we don't do this for Unipig. I think it makes sense that we keep Unipig all together because it is a pretty tightly coupled system, but I imagine we'll eventually move to a place where for instance importing the client shouldn't give you aggregator code.

Another thought is _maybe_ we consider merging this, tagging it, and then do even more comprehensive deletions. For instance, maybe even deleting the whole `plasma` package in order to reduce clutter as we push towards a comprehensive release. I've become a "delete with impunity because nothing is ever lost on the internet (esp with git)" maximalist lol.

Anyway, I actually think that this is looking pretty good to be merged! Impressed you got all the tests passing so quickly with so many changes!!!

